PR #1840, opened 2018-03-29
Supporting UTF8-SSID

ChunshengZhao:
Known issues: "kismet.netxml" file is not supporting UTF8-SSID.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -83,11 +83,11 @@ SRC_AV		= airventriloquist-ng.c
 
 SRC_SSE_COMMON	= memory.c wpapsk.c
 LIBAC_CFLAGS = -DOLD_SSE_CORE=1
-SRC_ACLIB	= cpuid.c crypto.c common.c $(SSEC_INT) uniqueiv.c $(ASM_AC)
+SRC_ACLIB	= cpuid.c crypto.c common.c $(SSEC_INT) uniqueiv.c $(ASM_AC) verifyssid.c
 LIBAC	= libaclib.la
 LIBAC_LIBS = libaclib.la
 
-AC_COMMON = cpuid.c crypto.c common.c $(SRC_SSE_COMMON) uniqueiv.c $(ASM_AC)
+AC_COMMON = cpuid.c crypto.c common.c $(SRC_SSE_COMMON) uniqueiv.c $(ASM_AC) verifyssid.c
 
 if LIBGCRYPT
 SRC_AC		+= sha1-git.c
@@ -401,7 +401,9 @@ EXTRA_DIST = wpaclean.c \
              mcs_index_rates.c \
              mcs_index_rates.h \
              cowpatty.c \
-             cowpatty.h 
+             cowpatty.h \
+             verifyssid.c \
+             verifyssid.h 
 
 
 @CODE_COVERAGE_RULES@
--- a/src/aireplay-ng.c
+++ b/src/aireplay-ng.c
@@ -72,6 +72,7 @@
 #include "osdep/osdep.h"
 #include "crypto.h"
 #include "common.h"
+#include "verifyssid.h"
 
 #define RTC_RESOLUTION  8192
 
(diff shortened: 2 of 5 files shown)


ChunshengZhao:
Just for aireplay-ng and airodump-ng, other programs also need to be modified. But I'm not sure if that will happen, because I don't know much about it.

jbenden (changes requested):
Inside of `verifyssid`, the `for` loop never performs checks against the length of `s`. This will cause buffer over-runs.

ChunshengZhao:
@jbenden 
Sorry, I rechecked the code and found no possibility of overflow.
The last character of a normal string is definitely 0x00.

e.g.
{0xe4,0x00};

If "c" is a utf8-flag(0xe4), s[i] is 0x00. Return 0.

jbenden:
The problem isn't what you're thinking, but rather that the string length is never tested and you've used `s[i+1]` a lot. It could read an invalid position of memory.

ChunshengZhao:
eg

jbenden:
Hello,

What I was proposing was to simply test lengths, too; like this:

```
int verifyssid(const unsigned char *s)
{
    int i;
    unsigned char c;
    int l = 0;

    if( !s || (l = (int) strlen((const char*)s)) > 32 ){//32 characters
        return 0;
    }

    for (i = 0; i < l && (c = s[i++]) != 0;) {
        if ((c & 0x80) == 0) {//ascii flag
            if(c < 0x20 || c == 0x7f){
                return 0;
            }
        } else if ((c & 0xe0) == 0xc0) {//utf8 flag
            if (i+1 < l && (s[i++] & 0xc0) != 0x80) {
                return 0;
            }
        } else if ((c & 0xf0) == 0xe0) {//utf8 flag
            if ((i+1 < l && (s[i++] & 0xc0) != 0x80) || (i+1 < l && (s[i++] & 0xc0) != 0x80)) {
                return 0;
            }
        } else if ((c & 0xf8) == 0xf0) {//utf8 flag
            if ((i+1 < l && (s[i++] & 0xc0) != 0x80) || (i+1 < l && (s[i++] & 0xc0) != 0x80) || (i+1 < l && (s[i++] & 0xc0) != 0x80)) {
                return 0;
            }
        } else {
            return 0;
        }
    }
    return 1;
}
```

I did not check for function correctness, but did check this code with a few static analyzers.

ChunshengZhao:
If you want to detect the length of the string it should be written like this:
But I don't think it is necessary, because accidental '\0' or invalid utf8-char will immediately break of the code and avoid illegal memory access.

```
int verifyssid(const unsigned char *s)
{
	int i;
	unsigned char c;
	int l = 0;

	if (!s || (l = (int)strlen((const char*)s)) > 32) {//32 characters
		return 0;
	}

	for (i = 0; i < l && (c = s[i++]) != 0;) {
		if ((c & 0x80) == 0) {//ascii flag
			if (c < 0x20 || c == 0x7f) {
				return 0;
			}
		}
		else if ((c & 0xe0) == 0xc0) {//utf8 flag
			if (i <= l && (s[i++] & 0xc0) != 0x80) {
				return 0;
			}
		}
		else if ((c & 0xf0) == 0xe0) {//utf8 flag
			if ((i <= l && (s[i++] & 0xc0) != 0x80) || (i <= l && (s[i++] & 0xc0) != 0x80)) {
				return 0;
			}
		}
		else if ((c & 0xf8) == 0xf0) {//utf8 flag
			if ((i <= l && (s[i++] & 0xc0) != 0x80) || (i <= l && (s[i++] & 0xc0) != 0x80) || (i <= l && (s[i++] & 0xc0) != 0x80)) {
				return 0;
			}
		}
		else {
			return 0;
		}
	}
	return 1;
}

```

jbenden (approved):

MPX4132:
@ChunshengZhao Thanks for taking the time to debug, man, not many people bother.

PR #1845, opened 2018-04-02
[Aireplay] add -C flag to set channel

ChrisLundquist:
before:
```
aireplay-ng --deauth 10 -e "<redacted>"  wlp3s0
21:03:22  Waiting for beacon frame (ESSID: <redacted> ) on channel 7
Found BSSID "CA:FE:00:00:BA:BE" to given ESSID "<redacted>".
21:03:24  wlp3s0 is on channel 7, but the AP uses channel 10
```

after:
```
 ./aireplay-ng --deauth 10 -e "<redacted>"  -C 10 wlp3s0 
21:03:41  Waiting for beacon frame (ESSID: <redacted>) on channel 10
Found BSSID "CA:FE:00:00:BA:BE" to given ESSID "<redacted>".
```

### DIFF
--- a/src/aireplay-ng.c
+++ b/src/aireplay-ng.c
@@ -204,6 +204,7 @@ char usage[] =
 "\n"
 "  Miscellaneous options:\n"
 "\n"
+"      -C channel_number     : set interface to channel_number before attacking\n"
 "      -R                    : disable /dev/rtc usage\n"
 "      --ignore-negative-one : if the interface's channel can't be determined,\n"
 "                              ignore the mismatch, needed for unpatched cfg80211\n"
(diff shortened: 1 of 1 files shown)


psyborg55:
don't use capital, -c is for channel not -C

aircrack-ng:
@psyborg55, -c already exists in aireplay-ng

ChrisLundquist:
for clarification, `-c` exists, as something other than setting channel.

psyborg55:
use -ch or --channel switch

ChrisLundquist:
@psyborg55 why are you so passionate about this?

psyborg55:
there are -c and -C flags in airodump-ng that i both use, a need to run updated aireplay-ng after your commit could easily bring up some confusion

kcdtv:
> there are -c and -C flags in airodump-ng that i both use, a need to run updated aireplay-ng after your commit could easily bring up some confusion

Well, -a means ` -a                    : Filter unassociated clients` in airodump-ng and means `-a bssid  : set Access Point MAC address` in aireplay-ng and etc... etc... 
Isn't this argument irrelevant?

ChrisLundquist:
I agree that it would be *ideal* to have a consistent interface for common options. Given that we are pretty far from that state, I think the "add a nice feature for now" is probably the way to go.

It seems to me that we should probably wait for a "2.0" release before reworking/standardizing all the flags for all the different tools as it would likely break any automation.

aircrack-ng:
@ChrisLundquist has a good point. The other reason for not changing the options is that scripts and other tools rely on Aircrack-ng. Changing an option after a release would break scripts and tools.

Issue #1850, opened 2018-04-04
32-bit macOS build failure


ryandesign:
Hi, building aircrack-ng 1.2-rc5 32-bit (`-arch i386`) fails on macOS 10.12.6:

```
libtool: link: ccache /usr/bin/clang++ -Wall -Wno-unused-command-line-argument -O3 -fstack-protector-strong -Wno-array-bounds -masm=intel -pipe -Os -stdlib=libc++ -arch i386 -Wl,-headerpad_max_install_names -arch i386 -o aircrack-ng--generic aircrack_ng__generic-aircrack-ng.o linecount.o  -L/opt/local/lib -lz -lsqlite3 ./.libs/libaclib.a ./.libs/libptw.a -lpthread -lssl -lcrypto -ldl -lm -pthread
Undefined symbols for architecture i386:
  "_GLOBAL_OFFSET_TABLE_", referenced from:
      _shasse2_init in libaclib.a(sha1-sse2.o)
      _shasse2_ends in libaclib.a(sha1-sse2.o)
      _shasse2_data in libaclib.a(sha1-sse2.o)
ld: symbol(s) not found for architecture i386
clang: error: linker command failed with exit code 1 (use -v to see invocation)
gmake[2]: *** [Makefile:1637: aircrack-ng--generic] Error 1
```

Building 64-bit (`-arch x86_64`) succeeds.

aircrack-ng:
What were the configure and make commands you ran?

jbenden:
It is questionable to even support 32-bit macOS. All of the underlying system is full 64-bit for quite some time...

jbenden:
PS: This would require assembly modifications in `src/sha1-sse2.S`. There is no entry for a 32-bit macOS PIC.

Also, not all x86 chips have SSE2; however, it could be a requirement for this case.

ryandesign:
Not all users are able to afford recent Macs, and some users may wish to repurpose older Macs. The first models of Intel Macs had 32-bit Intel Core CPUs.

1.2-rc4 built fine for 32-bit.

As far as I know, all Apple products with Intel processors support SSE2. The oldest Intel chip I'm aware of them using was the Pentium M in the Apple TV 1.

Build failures are never a good experience for the user, so I wanted to make sure you were aware of this one. If you don't want to support 32-bit, that's up to you, but maybe it could be detected at configure time and a more user-friendly error message could be printed.

jbenden:
[...]

PR #1853, opened 2018-04-06
32-bit macOS build failure

jbenden:
Fix for #1850


aircrack-ng:
This can be tested on 64 bit with `env CFLAGS="-arch i386" CXXFLAGS="-arch i386" ./configure`

Issue #1044, opened 2018-03-10
Monitor mode support on OSX


aircrack-ng:
**Reported by misterx on 25 Oct 2012 23:37 UTC**

I received a patch to add support for Monitor mode to Aircrack-ng on OSX. It requires the framework fundation. I'll test it soon.

aircrack-ng:
**Attachment by misterx on 25 Oct 2012 23:37 UTC**

Uploaded file [aircrack-ng_support_osx.diff](https://github.com/aircrack-ng/trac-attachments/blob/master/1044/20121025-233730_000-aircrack-ng_support_osx.diff)

osdep patch.

aircrack-ng:
**Comment by anonymous on 25 Oct 2012 23:37 UTC**

how to use this patch please? I install aircrack-ng it's macport

PR #1855, opened 2018-04-09
[Aireplay] support darwin

ChrisLundquist:
So this "works" as a proof of concept.
- Doesn't honor interface, just gets the default
- Super slow example code re-opens the interface over and over each time you read/write a packet.
- Fix build stuff like automake/autoconf

```
$ sudo ./src/aireplay-ng --test en1
I am on darwin, trying to open the wifi interface
23:13:15  Trying broadcast probe requests...
23:13:18  No Answer...
23:13:18  Found 1 AP 

23:13:18  Trying directed probe requests...
23:13:18  CA:FE:00:00:BA:BE - channel: 157 - '<redacted>'
23:13:35   0/30:   0%
```

We're leveraging `libpcap` to read/write packets, and we still have to deal with the radiotap headers.

libpcap doesn't seem to have the ability to change the channel of the adapter or some of the other parts of the osdep interface.

There are Objective-C bindings to manage things like channel as show in:
https://github.com/unixpickle/JamWiFi

Docs for the OSX framework at:
https://developer.apple.com/documentation/corewlan

A quick search didn't turn up any C bindings. Doing FFI into objective C is an option, but less than ideal.


ChrisLundquist:
I now see https://github.com/aircrack-ng/aircrack-ng/issues/1044 has a patch attached to it. This offers another way. Going the `iface fd` route of the other patch probably offers some lower level control which I like, adding Objective C and `execve` into `airport` isn't my favorite though, but does at least implement channel changing.

aircrack-ng:
It doesn't work on High Sierra. Using the framework foundation functions is the way to go.

PR #1857, opened 2018-04-11
Reworked how both Svn and Git versioning are formatted.

jbenden:
Added support for formatting both Svn and Git versioning information.

### DIFF
--- a/build/dist_check.sh
+++ b/build/dist_check.sh
@@ -18,7 +18,7 @@ fi
 make
 make dist
 
-V=$(cat VERSION)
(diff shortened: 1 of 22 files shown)


aircrack-ng:
It fails compiling when there is no source control used such as when doing svn export.

jbenden:
I had added an "unknown" for that case in #e15a4a5. Does it still? What's the export command, so I can locally test and fix?

jbenden:
Nevermind, I reproduced the problem.

jbenden:
Fixed.

PR #1868, opened 2018-04-20
[airodump-ng.c] Repair obfuscated security types

ChunshengZhao:
[airodump-ng.c:dump_write_csv() && dump_write_kismet_csv()] Repair obfuscated security types

Bug:
https://forum.aircrack-ng.org/index.php/topic,1543.0.html

The information provided is from a CSV file. Here is my guess.

#define	STD_OPN	0x0001
#define	STD_WPA2 0x0008

If ap_cur->security = 0x0009 or 0x000b or 0x000d and more...

ap_cur->security & STD_WPA2
ap_cur->security & STD_OPN

return **WPA2 and OPN**.

It may be a new encryption method (or damaged packet?), but I did not find the relevant information.

Sorry my English again.

### DIFF
--- a/src/airodump-ng.c
+++ b/src/airodump-ng.c
@@ -4116,42 +4116,33 @@ int dump_write_csv( void )
                  ap_cur->max_speed );
 
         if( (ap_cur->security & (STD_OPN|STD_WEP|STD_WPA|STD_WPA2)) == 0) fprintf( G.f_txt, " " );
-        else
-        {
-            if( ap_cur->security & STD_WPA2 ) fprintf( G.f_txt, " WPA2" );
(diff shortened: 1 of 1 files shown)


binarymaster:
> return WPA2 and OPN.
> It may be a new encryption method (or damaged packet?), but I did not find the relevant information.

When AP is WPS-enabled, Open association is allowed (along with WPA and/or WPA2) to transfer unencrypted EAP packets between AP and client.

ChunshengZhao:
@binarymaster 
So, this is a normal return value, but there is an error in the logical interpretation?
If if -> if “else if”

binarymaster:
Probably.

Issue #1811, opened 2018-03-10
Mutiple targets support


aircrack-ng:
**Reported by lumixx on 1 Sep 2017 05:18 UTC**

Can we have a mlutiple target APs support within the aireplay-ng deauth mode

For example, I have mutiple target APs on channel 1, can we have: 

aireplay-ng -0 0 -a bssid1,bssid2,bssid3 -x 1024 -D wlan0

So it can send packets fast enough, while affect them almost simultaneously, by interlacing those packets really fast to achieve this goal.
Let's say: 
10 packets to bssid1,
then 10 packets to bssid2, 
then 10 packets to bssid3, 
loop.

PR #1898, opened 2018-05-28
[aireplay-ng] multi target beta for deauth

k-kaminskas:
#1811 [aireplay-ng] multi target beta for aireplay-ng deauth mode

### DIFF
--- a/src/aireplay-ng.c
+++ b/src/aireplay-ng.c
@@ -161,13 +161,14 @@ char usage[] =
 "\n"
 "  Replay options:\n"
 "\n"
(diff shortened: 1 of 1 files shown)


ChrisLundquist:
https://github.com/aircrack-ng/aircrack-ng/compare/master...ChrisLundquist:clundquist/aireplay-deauth-all?expand=1

I did something similar here. My approach was just to listen for beacons for a time frame, then deauth any clients on that network if I detected any, or fallback to the broadcast deauth if I didn't detect a specific client.

Looks like the code format hit us both

PR #1933, opened 2018-07-19
GPS coordinates to stations in NetXML

### LINKED ISSUE #1565: Add GPS coordinates to stations in NetXML
**Reported by misterx on 22 Feb 2015 21:07 UTC**

Currently, we only have GPS coordinates for AP. We need to record that data for clients too (And update NetXML output)

SemiSkimmedSteve:
These changes implement GPS for stations resolves #1565.
Also captures best power for station to be able to get nearest (best) gps location.
I've basically copied and reused the code for the APs GPS.
Hope everything ok, I'm a Github Noob so go easy!

### DIFF
--- a/src/airodump-ng.c
+++ b/src/airodump-ng.c
@@ -1682,6 +1682,7 @@ static int dump_add_packet(unsigned char *h80211,
 		st_cur->tlast = time(NULL);
 
 		st_cur->power = -1;
+		st_cur->best_power = -1;
 		st_cur->rate_to = -1;
 		st_cur->rate_from = -1;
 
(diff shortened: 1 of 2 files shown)


jbenden:
Thanks for your contributions.

I had to modify your changes slightly, before accepting them. There were two issues:

1. Extra trailing whitespace. Please be certain (typically by using `git diff`) that changes are free of them, before submitting to OSS projects.
2. Instead of using `sizeof(float) * 5`, simply use the form `sizeof(structure)`. Reviewing the commit above will show you how it was modified. The reason this form is preferred is it protects the source code down the road, from structure changes and mismatched sizes. It's a form of DRY principle.

Thank you for your contributions!
-Joe

PR #1936, opened 2018-08-03
aircrack-ng: Support for 802.11w

jbenden:
This PR permits the `aircrack-ng` tool to process and crack 802.11w.


Mister-X- (approved):

ZeroChaos-:
I think this needs a little more documentation.  Perhaps a basic explanation of what key we are cracking (PSK, PMK, etc)?  I don't think most people even know what 802.11w is.

jbenden:
Perhaps the source is documented enough, by way of the function names surrounding each step of cracking? eg: The PMK is as-is. The PTK is a bit different, and the MIC is completely different.

jbenden:
It might need a wiki note, though...

ZeroChaos-:
I disagree, "read the code" isn't documentation.  This should be put in a man page or something *at least*.

PR #1943, opened 2018-08-22
Fix build with mmx

ffontaine:
Commit 39387fc80f90f3a9ac9ef9f3aa32da5776a0721e removed mmx support
however aircrack-ng fails to build on platforms with mmx because an
error is raised if __MMX__ is defined.

Fixes:
 - http://autobuild.buildroot.net/results/b7362b69435e9ef6fb2aedc50743e88dbd7a5c72

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/src/aircrack-crypto/arch.h
+++ b/src/aircrack-crypto/arch.h
@@ -357,9 +357,6 @@
 #elif __SSE2__
 #define SIMD_COEF_32 4
 #define SIMD_COEF_64 2
-#elif __MMX__
-#define SIMD_COEF_32 2
-#define SIMD_COEF_64 1
 #endif
 
 /*
--- a/src/aircrack-crypto/memory.h
+++ b/src/aircrack-crypto/memory.h
@@ -70,9 +70,6 @@
 #elif __SSE2__
 #define SIMD_COEF_32 4
 #define SIMD_COEF_64 2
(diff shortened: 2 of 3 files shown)


aircrack-ng:
Is there any currently supported Linux distribution that still handles such CPU, assuming it's using P5 microarchitecture (based on the toolchain being 586)?

ffontaine:
This toolchain is still in the buildroot's autobuilders (http://autobuild.buildroot.net/toolchains/tarballs/br-i386-pentium-mmx-musl-2018.05.tar.bz2) but I suppose there is not a lot of users.

If you don't like removing this code from aircrack-ng, an other option would be disable aircrack-ng on CPUs with MMU support.

aircrack-ng:
It compiles and runs on FreeBSD on a Pentium MMX. The only SIMD that works is 'generic'.

I'd love to see a use case for such toolchain though.

ZeroChaos-:
gentoo officially supports i386, i486, and i586

aircrack-ng:
Is there any good tutorial on setting up a i586 version?

jbenden:
I've accepted the patch, as-is, because it shouldn't have failed to build; it should have just built a generic binary.

Thanks!
-Joe

PR #1946, opened 2018-08-31
Cleanup OpenSSL includes

ChrisLundquist:
Pushes our OpenSSL includes down into the C file so that our other utilities don’t have to import a bunch of headers. This is possible because we don’t expose any OpenSSL types in the header.
We also don’t actually use sys/types in this file.

We also had a hidden dep that got pulled in from this. Turns out we need Malloc and Free in ivstool. (Which pulling in OpenSSL headers gave us before)

### DIFF
--- a/src/crypto.c
+++ b/src/crypto.c
@@ -32,6 +32,17 @@
 
  */
 
+
+#ifdef USE_GCRYPT
+#include "gcrypt-openssl-wrapper.h"
+#include "sha1-git.h"
+#else
+#include <openssl/hmac.h>
+#include <openssl/sha.h>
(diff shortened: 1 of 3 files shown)


aircrack-ng:
Both Travis (all 4) and AppVeyor fail (immediately)

ChrisLundquist:
Huh, I’ll take a look tonight. Make clean seemed to work locally. I must
have missed something.

ChrisLundquist:
Just wanted to follow up here to see if there is anything else I can do.
The tests passed great for the lats commit, turns out I missed grcypt initially.

Issue #1947, opened 2018-08-31
Aircrack-ng segmentation fault with test.ivs


aircrack-ng:
Running `aircrack-ng -K test.ivs` segfaults, when using `test.ivs` from https://aircrack-ng.org/doku.php?id=aircrack-ng#sample_files_to_try

The issue is due to #1934.

aircrack-ng:
@ChrisLundquist could you check this out?

ChrisLundquist:
Yeah, I should be able to take a look tonight

jbenden:
Please add this as a test (which crashes) before fixing. We'll then catch this condition in the future, if it crops up again... :) Thanks.

ChrisLundquist:
it looks like `ap_cur->ivbuf` is just plain null, and I don't see it ever get loaded/allocated.

It is a little tricky to setup a watchpoint since it is a local variable.

I'm of two minds on how to tackle this. One is to fix it short term somehow and ensure the ivbuf gets loaded / doesn't get freed. The other is to merge `check_thread` and `read_thread` which is a known task, but a large undertaking.

```
(lldb) settings set -- target.run-args  "-X" "-K" "./test.ivs"
(lldb) run
Process 31147 launched: './aircrack-ng' (x86_64)
Opening ./test.ivs
Read 567298 packets.

   #  BSSID              ESSID                     Encryption

   1  00:11:95:91:78:8C                            WEP (566693 IVs)

Choosing first network as target.

Opening ./test.ivs
1 potential targetsease wait...

Process 31147 stopped
* thread #1, name = 'aircrack-ng', stop reason = signal SIGSEGV: invalid address (fault address: 0x0)
    frame #0: 0x00007ffff6776b1f libc.so.6`__memmove_avx_unaligned_erms at memmove-vec-unaligned-erms.S:249
(lldb) bt
* thread #1, name = 'aircrack-ng', stop reason = signal SIGSEGV: invalid address (fault address: 0x0)
  * frame #0: 0x00007ffff6776b1f libc.so.6`__memmove_avx_unaligned_erms at memmove-vec-unaligned-erms.S:249
    frame #1: 0x00000000004149cc aircrack-ng`update_ivbuf at aircrack-ng.c:3683
    frame #2: 0x0000000000411150 aircrack-ng`do_wep_crack1(B=0) at aircrack-ng.c:3753
[...]

PR #1954, opened 2018-09-15
Fixes for wep segfault

ChrisLundquist:
ensures we only call read_thread xor check_thread.
Factors out duplication and starts to bring them closer.

for https://github.com/aircrack-ng/aircrack-ng/issues/1947


ChrisLundquist:
ok, this fixes the segfault as reported in #1947 but there are some other issues that are related.

I have to `git revert 1221bef55a8aaffc0ea2558a164e928fcaa290f1` and resolve the conflict (doesn't really matter how from what I see).

then `./src/aircrack-ng -q -b 00:11:95:91:78:8C -K ./test/test.ivs` will work, but you'll sometimes see an extra bogus answer like this:
```
clundquist@nuc:~/aircrack-ng$ ./src/aircrack-ng -q -b 00:11:95:91:78:8C -K ./test/test.ivs
1 potential targets

KEY FOUND! [ AE:5B:7F:3A:03:D0:AF:9B:F6:8D:A5:E2:C7 ] 
	Decrypted correctly: 100%

clundquist@nuc:~/aircrack-ng$ ./src/aircrack-ng -q -b 00:11:95:91:78:8C -K ./test/test.ivs
1 potential targets

KEY FOUND! [ AE:5B:7F:3A:03:D0:AF:9B:F6:8D:A5:E2:C7 ] 
	Decrypted correctly: 100%

KEY FOUND! [ AE:5B:7F:3A:03:D0:AF:9B:F6:8D:A5:BB:CC ] 
	Decrypted correctly: 0%

```

Note, `-q -b ...` will make it use `read_thread`. Things will "work" using `check_thread` but without `-q` you'll only see the last answer, which is often the bogus one.

I suspect 1221bef55a8aaffc0ea2558a164e928fcaa290f1 is in fact more correct, and we have something like an "off by one" else where that 1221bef55a8aaffc0ea2558a164e928fcaa290f1 detects.

I think I've taken this about as far as I'm willing. the TL;DR is `read_thread` and `check_thread` have drifted substantially. `read_thread` seems to be more correct. Swapping to the avl_tree just surfaced how much these methods have drifted.

aircrack-ng:
Looks like you added `src/tags` file by mistake

ChrisLundquist:
I finally get why `check_thread` and `read_thread` both exist and are different.
In check_thread, the goal (of `check_thread`) was to read "just enough" to print the chooser, then reload the data with `read_thread`. This saved memory because otherwise we have to allocate a bunch of WEP attack buffers and IVs and all sorts of things that we may not even care about if we're doing a WPA attack, or vice versa.

ChrisLundquist:
This branch seems reasonably stable. It survives a stress test with ~200GB of data.
```
# ...
995891  FF:F9:11:5E:91:A3                            WPA (0 handshake)
995892  FF:F9:11:6E:4C:7E                            WEP (1 IVs)
995893  FF:F9:11:6E:FE:78                            WEP (1 IVs)
995894  FF:F9:11:BE:46:73                            WPA (0 handshake)
995895  FF:F9:24:0C:48:7E                            WEP (1 IVs)
```

With that many APs and clients, the hot loop becomes searching for a client on an AP:
```
2468         while (st_cur != NULL)
2469         {
2470             if (!memcmp(st_cur->stmac, stmac, 6)) break;
2471 
2472             st_prv = st_cur;
2473             st_cur = st_cur->next;
2474         }
```

jbenden:
In master @ 8d76f43 I've really fixed the pipes problem. Apparently, they cannot be closed at all; however, they can be released during cleanup.

I am unsure of why there are incorrect results displayed, but I do know it seems to be only the last byte that is incorrect. So however the result is returned might not be sending the value across a pipe correctly...

jbenden:
FWIW: It seems `key_found` is called when `opt.probability` is even zero. If I filter that out, it seems to work. However, in looking some; I see that the `bf_pipe` is used to send over possible answers where the last two bytes need to be bruteforced. Because of this, the pipe has "extra" content after a fully probable key is found - thus leading to extra outputs.... It may be that shutdown code is missing to skip processing for more keys....

ChrisLundquist:
So, I'm not entirely in love with this PR, but I do think it is an improvement in a few regards such as adding a test, not seg faulting, removing the global buffer, etc.

The part that I'm struggling with is that it is a significant design decision change, that I'm not entirely sure I have complete context on. As I mentioned before, from what I've gleaned, `check_thread` was originally a first pass to load ' just enough data' to show the chooser. 'read_thread' was a second pass to fill in the desired target.

The upside of this approach was a notably smaller memory footprint, the downside was making multiple passes, and increased complexity.

This PR now brings `check_thread` much closer to `read_thread` in that it loads and initializes all the data it finds. The upside here is that it is single pass, and we could get rid of `read_thread`. Another plus is that we could more easily support a "multi target" mode if we load all the required data in one pass. The downside is it used ~8GB of memory when parsing ~200GB of packets caps. (This would be lower if specified a target as we'd skip other nodes)

So yeah, this is something to consider. I'm not sure how else I'd fix the segfault, I'd have to start over from `master` if we wanted to keep the idea of `read_thread` and `check_thread`. I know previously we said we wanted to collapse them into one. Now that I have worked with them more, I have a better understanding of why they are here.

jbenden:
Given the memory usage (and the requirement that the project run on small devices) I'd say to turn the combination of both threads into a singular (taking a parameter for SCAN vs. READ). The parameter is then used to determine what all is loaded. At the end of this function, if the mode was SCAN then recursively call itself again with READ set.

How's that sound?

jbenden:
PS: Travis failed probably due to `test.ivs` or `test-aircrack-ng-0010.sh` being missing from the `EXTRA_DIST` define inside of the `Makefile.am` script.

ChrisLundquist:
I thought I added it to automake. Half the Travis builds passed.

aircrack-ng:
#1953 was merged

aircrack-ng:
The scanning is not always done, it is only done when BSSID and maybe some other info is not available. So, in some cases, it will never be run.

jbenden:
@ChrisLundquist did you wish to continue the battle, or should one of us pick up from here?

ChrisLundquist:
I probably can’t push on this much more in the near term. I had to get a
job. The funny part is that it was about build tools and cross compiling.
Your name came to mind for someone who knew what they were doing when it
came to cross compiling and static linking OpenSSL for arm32.

Anyway, I’m happy to help as I can, as available. I didn’t want to leave
you all hanging with a segfault having introduced it.

aircrack-ng:
@ChrisLundquist Thanks a lot for the help

ChrisLundquist:
> Given the memory usage (and the requirement that the project run on small devices) I'd say to turn the combination of both threads into a singular (taking a parameter for SCAN vs. READ). The parameter is then used to determine what all is loaded. At the end of this function, if the mode was SCAN then recursively call itself again with READ set.
> 
> How's that sound?

I'm not sure that will buy us too much. That's more or less what `check_thread` + `read_thread` aims to do. There is a subtly that took me a long time to finally get, and I don't quite understand it well enough to communicate it.

> The scanning is not always done, it is only done when BSSID and maybe some other info is not available. So, in some cases, it will never be run.

I think Mr.X is pointing at this subtly with this statement.

The task of summarizing vs loading is different "just enough" that it made sense to have two different functions. As Mr.X said, if the user passes `-b <mac>` or `-e <essid` `check_thread` isn't needed, and we can jump straight to `read_thread`. However on the flip side, if we don't know what we're looking for we either need to stat it and ask the user, or load it all, and ask the user.

@Mister-X- sure thing. If there is anything to wrap this up I'm happy to do it. (I'll fix the merge conflict shortly)

Mister-X-:
@ChrisLundquist I don't think there is anything else, unless @jbenden can think of anything.

At first, #1947 sounded like an easy fix (like a misplaced `free()` or missing initialization) and since you knew the AVL changes very well, it would be easy to find, but it turned out much more complex than expected.

ChrisLundquist:
Roger. If you don't want to go this route that's perfectly OK, and feel
free to close this PR.
If nothing else, I think we've shed a lot of light on the issue.

jbenden:
Thanks everyone!

@ChrisLundquist if you need build system hints, let me know. :)

@Mister-X- Should I create a new branch and aggregate all of this, plus the changes I feel would work best? It seems this is priority as we're a tad broken currently...

Mister-X-:
@jbenden  Go ahead

PR #1956, opened 2018-09-17
Merge check and read threads

jbenden:
This PR resolves the problems of #1947, #1953, and #1954 by merging the operations of `check_thread` and `read_thread` in to a singular thread that performs both operations. This PR additionally fixes a few problems seen during code reading.


jbenden:
A future change might be to always call `packet_reader_thread` with `PACKET_READER_READ_MODE`, and use the `PACKET_READER_CHECK_MODE` to simply determine an AP's BSSID to load, populating `opt.bssid` so that the later call using `PACKET_READER_READ_MODE` functions to load all data associated with the AP. This is a better way to permit AP selection without loading all data up front.

At least for now, the `PACKET_READER_CHECK_MODE` discards all unused `struct AP_info` data, once a selection is made.

jbenden:
From the above, it introduces the requirement to read the entire packet capture file, twice; when going through the path that uses `PACKET_CAPTURE_CHECK_MODE` to display the available APs (or auto-select the only one) to select from.

It might be a better idea to have the `PACKET_CAPTURE_CHECK_MODE` save all file offsets for the packets in to a linked-list. Then upon selecting the desired AP, it simply seeks to these offsets, reads it, and continues until the linked-list is empty. This change isn't too difficult, but does add at least another element to `struct AP_info`.

aircrack-ng:
Adding another element is fine.

jbenden:
Before ever heading down this implementation path: is it even really necessary?

How many times does one process simply huge files; to read the entire thing, to manually select a BSSID, and continue?

Is it possible in this case to simply stick with the use of a secondary method of acquiring the desired BSSID and running `aircrack-ng` with the `-b` option?

In other words, how important is it that this functionality use the least amount of run-time memory? It ends up still using a lot of CPU after to crack...

aircrack-ng:
Huge files are probably not used very often.

jbenden:
Ok, sounds good. I'll hold on it, for now.

jbenden:
Please review for a merge to `master`.

src/aircrack-ng.c
```diff
@@ -392,6 +396,211 @@ static void ac_aplist_free(void)
 	pthread_mutex_unlock(&mx_apl);
 }
 
+/**
+ * Release all unused AP base-stations stored in \a access_points, keeping
+ * only the specified \a ap_cur AP base-station.
+ *
+ * @param ap_cur The AP base-station to keep.
+ */
+static void ap_avl_release_unused(struct AP_info *ap_cur)
```
ChrisLundquist, line 405: I was thinking about this too and was going with a "mark and sweep" approach.
I think we have an `avl_tree` of "targets" that got created based on the marked ones from the entire list.
We should be able to free all of `access_points` once we've finished our two passes and only use "targets".

Both these ways to us to the same place. I think we could probably call `ac_list_free` once we populate targets, which is more or less what this does.

src/aircrack-ng.c
```diff
@@ -589,23 +792,6 @@ static void sighandler(int signum)
 	if (signum == SIGWINCH) erase_display(2);
 }
 
-static void eof_wait(int *eof_notified)
```
ChrisLundquist, line 592: 💯

src/aircrack-ng.c
```diff
@@ -1032,1096 +1218,661 @@ static int atomic_read(read_buf *rb, int fd, int len, void *buf)
...
+		for (i = 0; i < clearsize; i++) clear[i + (32 * j)] ^= body[4 + i];
+	}
 
-	memset(&pkh, 0, sizeof(struct pcap_pkthdr));
-	memset(&pfh, 0, sizeof(struct pcap_file_header));
+	if (k == 1)
+	{
+		if (ap_cur->ptw_clean == NULL)
+		{
+			ap_cur->ptw_clean = PTW_newattackstate();
```
ChrisLundquist, line 1257: Note these two will (seemingly) use up a lot of memory and previously we only did this in read_thread.
I'm not sure if there is a code path to reach this though in check_mode now.

ChrisLundquist:
First off, nice work!


> From the above, it introduces the requirement to read the entire packet capture file, twice; when going through the path that uses `PACKET_CAPTURE_CHECK_MODE` to display the available APs (or auto-select the only one) to select from.

Yep, that's how the previous approach worked when calling `check_thread` it would read everything, show the stats, then `read_thread` would read everything again and filter for bssid/essid

> How many times does one process simply huge files; to read the entire thing, to manually select a BSSID, and continue?

I only ever do this to peek at the data. How many IVs, handshakes, etc.
I probably would have been better served by writing a --stats thing in hindsight.

> In other words, how important is it that this functionality use the least amount of run-time memory? It ends up still using a lot of CPU after to crack...

It isn't super important to me. a high water mark of 8GB after reading 200GB of data seemed acceptable. That means we only kept 4-5% of the data.

PR #1959, opened 2018-09-21
go back to the two pass method to lower memory

ChrisLundquist:
Without this method I can't load 40GB pcaps with 32GB of memory (no swap).

This is more or less what we are doing before in the `read_thread` and `check_thread` days, but we've greatly benefitted from the work of unifying the two code paths.

The commenting of `else` at the bottom is the key. We can/should reindent it, but I wanted to gather thoughts.

This is more or less what @jbenden mentioned in the comments of the other PR.

Nice to haves that crossed my mind:
- replace st_cur linked list with avl_tree (this is still the hot instruction for large pcaps)
- after the user finishes check_thread, print a helpful "next time use `aircrack-ng -b <the mac they chose> to start faster`" or at least tell them the mac/essid as picking 312231 isn't terribly helpful in the prompt.
- right now `packet_reader__update_ap_info` switches on format. We should break out these a parsers as we shouldn't have to check the format for each packet within a file (as I think we have always been doing). This will get rid of some branching / help improve pipelining. (<-- This is possible now that we have this in one place!)

### DIFF
--- a/src/aircrack-ng.c
+++ b/src/aircrack-ng.c
@@ -1294,7 +1294,7 @@ static int calculate_wep_keystream(unsigned char *body,
  * @return Returns zero on success. Returns non-zero for an error (> zero)
  *         or exception (< zero).
  */
-static int packet_reader__update_ap_info(struct AP_info *ap_cur, int fmt, unsigned char *buffer, unsigned char *h80211, struct ivs2_pkthdr *ivs2, struct pcap_pkthdr *pkh)
+static int packet_reader__update_ap_info(struct AP_info *ap_cur, int fmt, unsigned char *buffer, unsigned char *h80211, struct ivs2_pkthdr *ivs2, struct pcap_pkthdr *pkh, packet_reader_t *me)
 {
 	assert(ap_cur != NULL);
 	assert(buffer != NULL);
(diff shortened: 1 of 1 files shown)


jbenden:
Please fix the CI failure and fix the else indent (do not do other formatting changes - I realize it borked, but wanted a whole commit fixing everything at some point).

After those are done, I'll merge this. Then those other items can become PRs to contend with. :)

aircrack-ng:
There can be multiple DLT in pcap-ng AFAIK but that's the exception

ChrisLundquist:
> There can be multiple DLT in pcap-ng AFAIK but that's the exception

Sure, I meant switching on *file* type like this:
```
 if (fmt == FORMAT_IVS)
   // ...
 else if (fmt == FORMAT_IVS2)
   // ... 
 else 
   // parse pcap
 ```
We just need to factor out the pcap parsing into a function like the others are, and then we won't have to pass `fmt` to `packet_reader__update_ap_info`

ChrisLundquist:
> After those are done, I'll merge this. Then those other items can become PRs to contend with. :)

Should be fixed up, builds are cooking now!

PR #1962, opened 2018-09-22
print a helpful breadcrumb when we print the chooser

ChrisLundquist:
This makes it easier for users to take the fast path / single pass
```
# When I don't know what networks there are in a pcap
$ ./src/aircrack-ng /srv/wifi/home-01.pcapdump
# ...
Index number of target network ? 675
Use ./src/aircrack-ng -b F8:1D:0F:01:88:08 to load this target faster
Opening /srv/wifi/home-01.pcapdump
Read 11246612 packets.

1 potential targets

essid: "<redacted>" bssid: "F8:1D:0F:01:88:08"
Please specify a dictionary (option -w).

# When I know the target
 $ ./src/aircrack-ng  /srv/wifi/home-01.pcapdump -b F8:1D:0F:01:88:08
Opening /srv/wifi/home-01.pcapdump
Read 11246612 packets.

1 potential targets

essid: "<redacted>" bssid: "F8:1D:0F:01:88:08"
Please specify a dictionary (option -w).
```

### DIFF
--- a/src/aircrack-ng.c
+++ b/src/aircrack-ng.c
@@ -5970,7 +5970,7 @@ int main(int argc, char *argv[])
 	{
 		if (!opt.is_quiet)
 		{
-			printf("Reading packets, please wait...\r");
+			printf("Generating summary for target selection, please wait...\n");
 			fflush(stdout);
 		}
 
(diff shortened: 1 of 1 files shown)


aircrack-ng:
The tip should only be displayed if it takes more than a few seconds to parse the capture file(s). Something like 2-3 seconds.

Please add a newline before `Opening /srv/wifi/home-01.pcapdump`.

What is the purpose of displaying the essid and bssid again after selecting the network as well as the amount of potential target (which seems incorrect in the first output considering there are at least 675 networks)?

ChrisLundquist:
> What is the purpose of displaying the essid and bssid again after selecting the network

Feedback for the user in case they misread the number, or so they could -e or -b .

> amount of potential target (which seems incorrect in the first output considering there are at least 675 networks)?

We don't have to print that, but it is the number of networks that matched BSSID or ESSID (for multi target support)

aircrack-ng:
Let's remove the ESSID/BSSID printout.

For potential target, just add a warning if there is more than 1 BSSID that match the ESSID provided (it's up to you if you want to provide the list of BSSID)

PR #1964, opened 2018-09-24
comprehensive json-output, record rx-power per interface

emente:
did a patch that adds very comprehensible json output (that dumps received power per interface, too) for some experiments i currently do. maybe you will find it useful.

changes:
-added dump_write_json() plus supporting glue code
-json also writes rx powers per interfaces, had to add some glue code too (like interface name array)
-also dump non-associated info into json
-changed unknown and unassociated strings to something more unique (some clown here is actually using those)
-json-export obeys --berlin (another clown here is pseudo-randomizing macs)
-powers[] array, pkt_cnt, manufacturer now exists in all ap,st,na structs
-upped MAX_CARDS and increased ring buffers slightly

missing: 
-reset powers[] arrays after channel wrap around (eg, if scanner moves)

### DIFF
--- a/manpages/airodump-ng.8
+++ b/manpages/airodump-ng.8
@@ -72,7 +72,7 @@ Display APs uptime obtained from its beacon timestamp.
 Display a WPS column with WPS version, config method(s), AP Setup Locked obtained from APs beacon or probe response (if any).
 .TP
 .I --output-format <formats>
-Define the formats to use (separated by a comma). Possible values are: pcap, ivs, csv, gps, kismet, netxml. The default values are: pcap, csv, kismet, kismet-newcore.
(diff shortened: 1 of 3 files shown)


aircrack-ng:
A few important notes:
- Please don't change the CSV format, it will break existing tools
- If there are issues to fix, open a separate PR
- `--berlin` option is a display option, so it doesn't affect output files, unless the output format is about providing updates

Initial review:
- Could you use comments instead of long strings of `/`?
- Format (`--output-format`) is missing from the help screen, parsing and the man page
- While JSON is valid, there are some weird values in the `powers` item when using a single adapter (which is supposed to represent the adapter name)
- GPS coordinates are not handled in the JSON

emente:
no time to review the collisions the reformatting in the master caused

PR #1980, opened 2018-10-20
[airodump-ng] Revamped the GPS logging functionality

nathanh2011:
This is my first contribution so please be gentle.

As a bit of background I have been using Aircrack (airodump specifically) for many years for WIFI survey and mapping. I have made the following changes several times over the years but thought I would offer it to the main branch to save me having to add it after every update and if other people can get use from it. Essentially I found the GPS logging and default file formats for recording actual locations of WIFI APs to be lacking and incorrect (especially if you lose GPS signal). The majority of my work is dedicated to accurately map signal strengths of APs so I modified the Airodump program to give me the details and resolution I needed. 

I hope someone else finds this helpful. Below are the actual changes.

Revamped the GPS logging functionality
- Changed the GPS logging function to be a thread instead of a separate process for closer integration
- Wrapped the GPSd socket reader in a while loop to reconnect if we lose connection to the service
- Fixed the GPS to only provide a valid fix and reset if no valid fix is found in 5 seconds (prevents incorrect GPS data being used)

Added a new output format 'logcsv'
- The logcsv will write a CSV line on each refresh to output the seen APs with basic info and the current GPS fix. This format is ideal for wardriving and gives a more accurate map of APs and signal strengths compared to the max/avg/min you get from the other formats.

### DIFF
--- a/src/airodump-ng.c
+++ b/src/airodump-ng.c
@@ -730,7 +730,7 @@ char usage[] =
 	"      --wps                 : Display WPS information (if any)\n"
 	"      --output-format\n"
 	"                  <formats> : Output format. Possible values:\n"
-	"                              pcap, ivs, csv, gps, kismet, netxml\n"
+	"                              pcap, ivs, csv, gps, kismet, netxml, logcsv\n"
 	"      --ignore-negative-one : Removes the message that says\n"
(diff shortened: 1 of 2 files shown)


jbenden:
Merged, with slight changes. Thanks!

PR #1982, opened 2018-10-25
ax_ext_have_static_lib.m4: allow overriding lib name

alonbl:
introduce AX_EXT_HAVE_STATIC_LIB_DETECT macro to detect the location of the
lib directories, allow to specify the lib component (lib, lib64) and use the
detected host instead of gcc explicit command.

the change requires converting the DEFAULT_STATIC_LIB_SEARCH_PATHS from m4
macro to environment variable.

Signed-off-by: Alon Bar-Lev <alon.barlev@gmail.com>

### DIFF
--- a/build/m4/aircrack_ng_crypto.m4
+++ b/build/m4/aircrack_ng_crypto.m4
@@ -45,8 +45,9 @@ AC_ARG_ENABLE(static-crypto,
     [static_crypto=$enableval], [static_crypto=no])
 
 if test "x$static_crypto" != "xno"; then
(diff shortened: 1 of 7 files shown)


jbenden:
Hello, thanks for the contribution!

Unfortunately, this PR breaks the existing packaging infrastructure; as all OSes in use, use the `gcc` reported output as the path component inside of `/usr/lib`.

This PR uses: `/usr/lib/x86_64-pc-linux-gnu`
But the OS uses: `/usr/lib/x86_64-linux-gnu`

Could this be modified to also (or only) use the `gcc` reported path?

Thanks,
-Joe

alonbl:
Hi!
Interesting... I thought that gcc always has a matching $host.
I pushed another version of the patch which includes gcc specific code.
Thanks!

PR #1991, opened 2018-11-16
Fix a few warnings

neheb:


neheb:
One benefit is that the dependency on gnu99 can now be removed.

aircrack-ng:
It currently breaks compilation

neheb:
Yeah the second commit is faulty. I'll remove it soon.

aircrack-ng:
Still breaks on Travis and AppVeyor

src/aircrack-util/crypto_engine_loader.c
```diff
@@ -198,7 +198,7 @@ void (*dso_ac_crypto_engine_calc_mic)(ac_crypto_engine_t * engine,
 	= NULL;
 #endif
 
-#if defined(CYGWIN)
+#if defined(__CYGWIN__)
```
jbenden: This macro is defined via the Autoconf scripts.

jbenden (commented):
Uck!

src/aircrack-util/crypto_engine_loader.c
neheb: Yeah it was an attempt to fix the buildbots. Apparently the reason they fail is that they pass -Werror, which is fairly amusing.

Speaking of which, the cygwin defines are all over the place.

jbenden: Glad you like that. :)

jbenden (changes requested):
Please look at the quantity of files/lines changed with the diff. I'm sure it broke Windows line endings on certain files...

Could you revert to an older commit and try that change over?

neheb:
Yeah I'll fix it. I accidentally added unrelated stuff.

neheb:
Rebased. Single buildbot failure seems to be due to some license thing.

jbenden (approved):

PR #2000, opened 2018-12-14
APs and clients in  'logcsv'

r4ulcl:
Modification of the new output format 'logcsv' to log APs and clients.


aircrack-ng:
Like this, if anybody implements a tool to parse the current version, this will break it. Could you put the `type` at the end of it, so breakage wouldn't happen (if the parser splits on comma and uses position).

Looking at the other items, each of them has its first character uppercase, so could you do the same for `Type`

aircrack-ng:
Thanks

Issue #1958, opened 2018-09-20
WPAclean: keep timestamps in output files


aircrack-ng:
Looks like wpaclean uses zero timestamps in its output file(s) according to #1937.

Keep the original timestamps when writing files.

jbenden:
I took a quick look. It seems that `wi_read()` from osdep does not make the timestamp information available (from looking at `file.c`'s `read_file` implementation); so all of osdep would need changed to support accessing this data.

ZerBea:
Keeping timestamp will be another big improvement for aircrack-ng.
Now we can determine a good handshake based on measured values of the EAPOL timer of the access point.

aircrack-ng:
Combining with #1903, the first frame might get a 0 timestamp but there will be time difference in-between frames

ZerBea:
BTW:
Another big improvement would be to exclude deauthentication frames. This frames are mostly useles for an analyze.
This is a typical aircrack-ng cap file:
1 beacon and many deauthentication frames.
Please consider (for exampel for besside-ng) to save authentication, association (request and response), re-association (request and response) and EAP frames, too. They include some nice informations. I can send you some cap files....

beacons (with ESSID inside)..: 1
deauthentications............: 222657

$ hcxpcaptool -V 20180920_201711_80.cap
start reading from 20180920_201711_80.cap
summary:                                        
file name....................: 20180920_201711_80.cap
file type....................: pcap 2.4
file hardware information....: unknown
file os information..........: unknown
file application information.: unknown
network type.................: DLT_IEEE802_11 (105)
endianess....................: little endian
read errors..................: flawless
packets inside...............: 291451
skipped packets..............: 0
packets with FCS.............: 0
beacons (with ESSID inside)..: 1
probe responses..............: 584
association requests.........: 2
association responses........: 18
reassociation responses......: 1
authentications (OPEN SYSTEM): 32
authentications (BROADCOM)...: 11
deauthentications............: 222657
disassociations..............: 92
EAPOL packets................: 618
best handshakes..............: 1 (ap-less: 0)

ZerBea:
The dumper (airodump-ng, besside,ng, ...) can get the timestamp from the socket:
ioctl(fd_socket, SIOCGSTAMP , &tv)
tv is a timeval struct
I saw the feature request to add pcapng support. Now you can convert the timval to pcapng format:
timestamp = (tv.tv_sec * 1000000) + tv.tv_usec;
[...]

PR #2010, opened 2019-01-05
De-duplicate functions; with other minor tweaks

jbenden:


jbenden:
Re #1958; we would now output the timestamps (as is) from the input pcap files.

PR #2016, opened 2019-01-21
Build cleanups

alonbl:
Introduce some build cleanups.

1. Fix tests to work under `make distcheck`
2. Remove libtool archives from sources as incorrect and blocks `make dist`
3. Use libtool archives instead of explicit flags to follow libtool best practice

After this set the `make dist` and `make distcheck` finally works, and you are able to publish a distribution tarball to help upstream to avoid having to generate the build with possibility incompatible environment.


alonbl:
@jbenden : thanks! will you add the `make dist` tarball to next release as an artifact?

jbenden:
I believe there are problems with doing so, in a portable manner, across all supported platforms. I had tried long ago and it wasn't working.

If you could test this across at least two platforms, I'd reconsider. Or if you strongly believe it would work, I'll schedule some time to test myself.

Thanks,
-Joe

alonbl:
Based on my experience it is the opposite :)

The package is self contained and in worse case scenario autoreconf should be run again.

I am the Gentoo maintainer of this package, using the distribution tarball will make our lives much easier, Once you release, I will modify builds to use the distribution package and help you to resolve issues if happens.

PR #2030, opened 2019-03-07
Fix wpaclean to write all, and not only the last captured handshake to the output file.

buzzdeee:
In save_network(), open_pcap() is called for each detected network
in the input capture files. open_pcap() truncates the existing
file, and starts from zero with each network it finds.
So only call open_pcap() on the first network/handshake it detects,
so that subsequtent handshakes are added to the output capture file.

### DIFF
--- a/src/wpaclean.c
+++ b/src/wpaclean.c
@@ -149,7 +149,10 @@ static void save_network(const struct timespec * ts, const struct network * n)
 
 	int i;
 
-	_outfd = open_pcap(_outfilename);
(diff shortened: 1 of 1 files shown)


jbenden:
This pull request has been rebased via `git hub pull rebase`. Original pull request HEAD was e35996306a0e001495a367a2adc8a42319328087, new (rebased) HEAD is 2bf1dcb45f27d557d4ef3552fcc928701a0a4e85

PR #2039, opened 2019-04-16
autoconf: fix openssl static library order

rofl0r:
if installed only as static libs, the order of the link command line matters,
and -lssl requires symbols which -lcrypto provides.
the second spot in this file had it right.
(it doesn't matter with dynamic libs due to DT_DYN tab containing all
required library names per shared object).

### DIFF
--- a/build/m4/ax_check_openssl.m4
+++ b/build/m4/ax_check_openssl.m4
@@ -50,7 +50,7 @@ AC_DEFUN([AX_CHECK_OPENSSL], [
             esac
         ], [
             AC_CHECK_LIB([crypto], [OPENSSL_init], [
-                OPENSSL_LIBS="-lcrypto -lssl"
+                OPENSSL_LIBS="-lssl -lcrypto"
                 OPENSSL_LDFLAGS=""
 
                 AC_CHECK_HEADERS([openssl/crypto.h], [


rofl0r:
the appveyor build failure seems to be due to a wrong cygwin version isntalled:

```
      1 [main] bash (612) C:\cygwin\bin\bash.exe: *** fatal error - cygheap base mismatch detected - 0x13D2410/0xF22410.
This problem is probably due to using incompatible versions of the cygwin DLL.
Search for cygwin1.dll using the Windows Start->Find/Search facility
```

jbenden:
I've restarted the failed sub-builds.

AppVeyor is notoriously flakey....

rofl0r:
hmm, now the build and the tests passed, yet it is still listed as fail...

rofl0r:
so everything seems fine, right ? as the other spot in the file already uses -lssl -lcrypto, merging this should be a no-brainer

rofl0r:
ping @jbenden

rofl0r:
ping @aircrack-ng

DJDan:
@jbenden @aircrack-ng 
Since this commit have already been successfully passed can this please get merged back to the main aircrack-ng source code? That way I dont need to keep manually editing the file to cross compile. Thanks.

PR #2051, opened 2019-05-24
Add support for the rtl8188eus driver stack

kimocoder:
* Realtek base v5.3.9 with monitor mode & frame injection, supporting lots of cards.

### DIFF
--- a/scripts/airmon-ng.linux
+++ b/scripts/airmon-ng.linux
@@ -806,9 +806,7 @@ getFrom() {
 				FROM="V"
 			elif [ "$DRIVER" = "rt5390sta" ]; then
 				FROM="V"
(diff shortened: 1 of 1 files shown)


scripts/airmon-ng.linux
```diff
@@ -805,6 +808,8 @@ getFrom() {
 				FROM="V"
 			elif [ "${DRIVER}" = "8812au" ] || [ "${DRIVER}" = "8814au" ]; then
 				FROM="V"
+			elif [ "${DRIVER}" = "rtl8188eus" ] || [ "${DRIVER}" = "rtl8188eus" ]; then
```
ZeroChaos-: why the same thing twice?

kimocoder: Aye, sorry. I'll fix it right away

PR #2055, opened 2019-06-01
Prevents deprecated drivers from entering findFreeInterface function

CarterFendley:
`findFreeInterface` function uses the `iw` command to add new interfaces. Because `iw` supports only mac80211 drivers, unexpected behavior could happen when deprecated drivers are sent to this function. 

The commit moves the check for deprecated drivers above the statement checking for interfaces with long names. As the `startDeprecatedIface` doesn't append "mon" to the end, the check of name length does not seem to be necessary for deprecated drivers.

### DIFF
--- a/scripts/airmon-ng.linux
+++ b/scripts/airmon-ng.linux
@@ -443,6 +443,14 @@ startMac80211Iface() {
 			fi
 		done
 	fi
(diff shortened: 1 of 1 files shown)


ZeroChaos-:
this change is fine, although next time please try to be a bit nicer to the whitespace :-)

thanks!

PR #2058, opened 2019-06-08
Added support for dirty rtl8188eus chipset

kimocoder:
* Added support for dirty (nl80211/cfg80211 mixed) rtl8188eus drivers
* Supports lots of adapters
* Monitor mode & frame injection supported


ZeroChaos-:
Please don't remove 8814, just because it isn't current doesn't mean no one has it installed

PR #2066, opened 2019-07-25
Shellcheck warnings

neheb:


ZeroChaos-:
Honestly it ruins readability to change the way the printfs are, and I haven't found a shell that can't handle the current syntax.  NACK

ZeroChaos-:
okay, so I mostly committed the real fixes, everything but the %s stuff and pgrep.  thanks a bunch for finding that stuff.  I will take a look at if openwrt has pgrep by default, but if so I'll merge that too.

neheb:
Current SNAPSHOT does. Not sure about previous releases.

edit: The current supported releases (18 and 19 do. 17 should as well.)

PR #2073, opened 2019-08-21
[airdecap-ng] Decrypt both directions of WDS

jacobopantoja:
BSSID and STA in WDS mode can only be properly identified if BSSID is pre-known. Therefore, we use the BSSID parameter in the command line to properly identify STA and BSSID. If no BSSID is provided, previous behaviour is reproduced (only traffic destined to AP is decrypted), and a warning is given.

### DIFF
--- a/src/airdecap-ng/airdecap-ng.c
+++ b/src/airdecap-ng/airdecap-ng.c
@@ -77,6 +77,9 @@ static const char usage[] =
 	"      -k <pmk>   : WPA Pairwise Master Key in hex\n"
 	"\n"
 	"      --help     : Displays this usage screen\n"
+	"\n"
+	"  If your capture contains any WDS packet, you must specify the -b\n"
+	"  option (otherwise only packets destined to the AP will be decrypted)\n"
 	"\n";
 
 static struct decap_stats
@@ -109,6 +112,7 @@ static struct options
 
 static unsigned char buffer[65536];
 static unsigned char buffer2[65536];
+static bool wds = false;
 
 /* this routine handles to 802.11 to Ethernet translation */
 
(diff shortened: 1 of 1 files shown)


aircrack-ng:
Would you be able to provide test data?

jacobopantoja:
> Would you be able to provide test data?

Yes, sorry. Here is the original capture (captured with `airodump-ng --channel 140 -w capture_wds --output-format pcap wlp1s0mon`), plus decrypted with Arch Linux airdecap-ng 1.5.2, plus decrypted with current master modified with my patch.
Both AP and STA are OpenWRT boxes with 4-address set (so to make a wireless bridge), containing VLANs. I made tests with and without VLANs to ensure there was no influence, and results are identical regardless of the VLANs.

SSID: `test1`
Passphrase: `12345678`

**capture-wds-01.cap** -> Original capture

**capture-wds-01-dec_vanilla_1.5.2.cap** -> Decrypted with vanilla 1.5.2, with `airdecap-ng /tmp/capture_wds-01.cap -e test1 -p 12345678 -l`
   43 out of 46 packets are decrypted (the 3 remaining are from the AP to the STA). Additionally, I'm seeing that in the stats, `Total number of stations seen` is 2 (I guess should be only 1).

**capture-wds-01-dec-patched_1.5.2.cap** -> Decrypted with patched 1.5.2, with `airdecap-ng /tmp/capture_wds-01.cap -e test1 -p 12345678 -l -b 00:11:22:00:00:00`
   46 out of 46 packets are decrypted, and `Total number of stations seen` is 1. If decrypted without `-b`, results are identical to the vanilla ones.

Looking forward for your comments.

[captures.zip](https://github.com/aircrack-ng/aircrack-ng/files/3528603/captures.zip)

PR #2080, opened 2019-09-06
Refactoring of airodump-ng and other related files

ChrisNisbet01:
This is a fairly significant change, so I appreciate that you may be reluctant to take this without some serious thought and/or testing. I'm just bringing it to your attention. Do with it what you will.
Note that I had added some UBUS support and another output format for my own purposes, but removed all of that in this branch so that it more closely follows the support in your project.
The general idea from my perspective was to (hopefully) make the code a little clearer and modular.
Everything still builds on my local machine, and airodump-ng still appears to work, but I haven't checked any of the other applications to ensure they're still operational; I wouldn't know what they're used for anyway.
I haven't made any attempt to follow your own coding standards, but I figure that if you're interested in these changes, it wouldn't take too much effort to run a formatter over them.
I know that at least a couple of the new headers files are probably in the wrong place.


aircrack-ng:
After a quick check, here are the initial remarks:
- The OUI file is not mandatory and there is a tool to download it
- There are a lot of unrelated commits, it would be preferable if this had just the related commits, to make it easier to review them
- You removed a lot of "const" from the code, why is that?
- Are the .gitignore changes all necessary?
- Style can be checked easily, we have a .clang-format file, that can be loaded with a lot of code editors or just used in the command line with clang-format. Ideally, with clang-3.9
- Travis noted an error, and it feels like it's just a missing header in one of the Makefiles:

```
/home/travis/build/aircrack-ng/aircrack-ng/dist_build/aircrack-ng-1.5.2_rev-b9ea99c/include/aircrack-ng/support/communications.h:51:42: fatal error: aircrack-ng/osdep/mac_header.h: No such file or directory
```

ChrisNisbet01:
I have removed airodump-ng-oui.txt from the repo and removed some of those changes to .gitignore.
I think you'll find I have _added_ a lot of consts to the code. You would need to point me to the consts you think I have removed.
Yes, there are a lot of commits, but they are all related.
I really don't know enough about the make system to know if I was making the appropriate changes to get all the required headers into the tarball, which is where I think things went wrong.
If these changes are interesting to anybody, they might like to lend a hand and resolve some of the issues.

Issue #2019, opened 2019-02-04
error installing: Undefined symbols for architecture x86_64


astonly:
S-MacBook-Pro:aircrack-ng Stas$ autoreconf -i
S-MacBook-Pro:aircrack-ng Stas$ ./configure --with-experimental
checking build system type... x86_64-apple-darwin18.2.0
checking host system type... x86_64-apple-darwin18.2.0
checking target system type... x86_64-apple-darwin18.2.0
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... ./install-sh -c -d
checking for gawk... no
checking for mawk... no
checking for nawk... no
checking for awk... awk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking whether to enable maintainer-specific portions of Makefiles... yes
checking whether make supports nested variables... (cached) yes
checking whether make supports the include directive... yes (GNU style)
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking whether gcc understands -c and -o together... yes
checking dependency style of gcc... gcc3
checking for g++... g++
checking whether we are using the GNU C++ compiler... yes
checking whether g++ accepts -g... yes
checking dependency style of g++... gcc3
checking dependency style of gcc... gcc3
checking for pkg-config... /usr/local/bin/pkg-config
checking pkg-config is at least version 0.9.0... yes
checking how to print strings... printf
checking for a sed that does not truncate output... /usr/bin/sed
checking for grep that handles long lines and -e... /usr/bin/grep
checking for egrep... /usr/bin/grep -E
checking for fgrep... /usr/bin/grep -F
checking for ld used by gcc... /Library/Developer/CommandLineTools/usr/bin/ld
checking if the linker (/Library/Developer/CommandLineTools/usr/bin/ld) is GNU ld... no
checking for BSD- or MS-compatible name lister (nm)... /usr/bin/nm -B
checking the name lister (/usr/bin/nm -B) interface... BSD nm
checking whether ln -s works... yes
checking the maximum length of command line arguments... 196608
[...]

PR #2088, opened 2019-10-05
airodump-ng: bug fix colorize option

scorpius:


aircrack-ng:
Could you give more details: when it happens and how to reproduce the bug

scorpius:
If you activate the feature by pressing the o key, all clients have the same color. But with the fix, each client matches the color of the AP it is associated to.

aircrack-ng:
With this patch, existing clients not linked to AP in the top section. This is most likely due to a bug in the auto-coloring - See #2019

aircrack-ng:
With this patch, existing clients not linked to AP in the top section become hidden when hitting 'o'. This could be due to a bug in the auto-coloring - See #2019

scorpius:
On my system the non-linked-to-ap clients are visible and have a grey color. I think it is not a bug, but different terminal colors can make it look hidden.

PR #2090, opened 2019-10-22
Replace ps | grep with a simple pgrep

neheb:
Eliminates busybox hacks and allows it to work properly with busybox.

### DIFF
--- a/scripts/airmon-ng.linux
+++ b/scripts/airmon-ng.linux
@@ -1088,22 +1088,6 @@ getExtendedInfo() {
 }
 
 scanProcesses() {
-	#this test means it errored and said it was busybox since busybox doesn't print without error
-	if (ps -A 2>&1 | grep -q BusyBox)
-	then
-		#busybox in openwrt cannot handle -A but its output by default is -A
-		psopts=""
-	else
-		psopts="-A"
-	fi
-	if ( ps -o comm= 2>&1 | grep -q BusyBox )
-	then
-		#busybox in openwrt cannot handle -o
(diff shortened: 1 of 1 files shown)


scripts/airmon-ng.linux
```diff
@@ -1120,12 +1104,7 @@ scanProcesses() {
 	fi
 
 	unset match
-	if [ "${pso}" = 1 ]; then
-		match="$(ps ${psopts} -o comm= | grep ${PROCESSES} | grep -v grep | wc -l)"
-	elif [ "${pso}" = 0 ]; then
-		#openwrt busybox grep hits on itself so we -v it out
-		match="$(ps ${psopts} | grep ${PROCESSES} | grep -v grep | wc -l)"
-	fi
+	match="$(pgrep ${PROCESSES} | grep -c -v grep)"
```
neheb: Actually the grep here seems like it can be replaced by wc -l.

aircrack-ng:
Doesn't work on Kali (and probably other distros that don't use busybox). Output is empty with a patched airmon-ng, and reports processes on unpatched.

jbenden:
I may be wrong, but I believe that `pgrep` is not available across the wide range of supported operating systems. That said, for Debian-based distros, the `pgrep` tool is provided by the `procps` package.

neheb:
I see the problem... pgrep can only handle one expression at a time. It does not support the \| syntax used in PROCESSES . Meaning a loop would need to be created...

I doubt that would simplify the script. Closing.

Issue #2062, opened 2019-06-29
Error in airodump-ng: double free or corruption


ghost:
# Issue type

- Defect - Crash
```
ioctl(SIOCSIWMODE) failed: Invalid argument
ioctl(SIOCSIWMODE) failed: Invalid argument
Error setting monitor mode on enp0s29f7u8
*** Error in `airodump-ng': double free or corruption (!prev): 0x0000000000d15cf0 ***
```

# System information

- OS: `Linux centos 3.10.0-957.21.3.el7.x86_64 #1 SMP Tue Jun 18 16:35:19 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux`
- CPU: `Intel(R) Pentium(R) Dual  CPU  E2160  @ 1.80GHz`
- Wireless card and chipset (if relevant): `Bus 001 Device 003: ID 0bda:8812 Realtek Semiconductor Corp. RTL8812AU 802.11a/b/g/n/ac 2T2R DB WLAN Adapter`

# Aircrack-ng version

- Version: `Aircrack-ng 1.5.2  - (C) 2006-2018 Thomas d'Otreppe`
- Commit Revision hash: downloaded from `https://download.aircrack-ng.org/aircrack-ng-1.5.2.tar.gz`

# Defect
## How to reproduce the issue

After following these instructions: `https://www.aircrack-ng.org/doku.php?id=install_aircrack#installing_aircrack-ng_from_source`, when I run `airodump-ng enp0s29f7u8`, I have this error:
```
ioctl(SIOCSIWMODE) failed: Invalid argument
ioctl(SIOCSIWMODE) failed: Invalid argument
Error setting monitor mode on enp0s29f7u8
*** Error in `airodump-ng': double free or corruption (!prev): 0x0000000000d15cf0 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x81489)[0x7fb3fe817489]
/usr/local/lib/libaircrack-osdep-1.3.0.so(+0x7377)[0x7fb3ff4d2377]
/usr/local/lib/libaircrack-osdep-1.3.0.so(wi_open+0x68)[0x7fb3ff4cd7d8]
airodump-ng[0x403c3b]
/lib64/libc.so.6(__libc_start_main+0xf5)[0x7fb3fe7b83d5]
airodump-ng[0x407464]
======= Memory map: ========
00400000-00420000 r-xp 00000000 fd:00 100793962                          /usr/local/sbin/airodump-ng
[...]

PR #2094, opened 2019-11-10
fix for crash (double free) when libnl3-devel (for HT channel support…

dmfreemon:
…) is not installed (fedora).

The configure script unsets CONFIG_LIBNL, which leads to the double free.


dmfreemon:
Looking at the Travis output, I don't see what the problem might be.  make check works fine on my machine.  Please advise if there is something I should do for this.  Thanks.

jbenden:
It may have been an intermittent thing, on Travis's part. The macOS image hung with no response after 10 minutes. I've restarted the job.

dmfreemon:
Looking at the Travis output, I'm not sure I know what the problem is.  make check works fine on my machine.  But my machine is linux, and I think the build failure is on osx.

I do see this in Travis:
> checking whether build environment is sane... configure: error: newly created file is older than distributed files!
> Check your system clock

It's not clear if that is causing the non-zero return code.  But I am wondering if this could be indicating a problem with the build machine.

mgaflip:
see https://github.com/aircrack-ng/aircrack-ng/issues/2062#issuecomment-597567304

lib/osdep/linux.c
```diff
@@ -2259,7 +2257,7 @@ static int do_linux_open(struct wif * wi, char * iface)
 	dev->arptype_in = dev->arptype_out;
 
 	if (iface_malloced) free(iface);
-	if (iwpriv) free(iwpriv);
```
mgaflip, line 2262: this line should be deleted/comented as iwpriv points to a static char array

lib/osdep/linux.c
```diff
@@ -2275,7 +2273,6 @@ static int do_linux_open(struct wif * wi, char * iface)
 	close(dev->fd_in);
 	if (acpi) fclose(acpi);
 	if (iface_malloced) free(iface);
-	if (iwpriv) free(iwpriv);
```
mgaflip, line 2278: this line should be deleted/commented as iwpriv points to a static char array

dmfreemon:
@mgaflip Thanks for taking the time to share your comments above.  Can you confirm that you were _not_ using this PR when experiencing this crash and that your code change suggestions above are referring to the current master code and not code in this PR?

I believe the existing code in this PR has already fixed this bug.

mcampana:
Hi, I'm using the master code. The big is there

mcampana:
Bug, I mean

lib/osdep/linux.c
RafaelLeeImg: This line should be deleted/comented, agreed.
The iwpriv will not exist anymore after this patch merges. All stuff return by wiToolsPath should be free, wiToolsPath  uses malloc inside.

lib/osdep/linux.c
RafaelLeeImg: Same as above.
This line should be deleted/comented, agreed.
The iwpriv will not exist anymore after this patch merges. All stuff return by wiToolsPath should be free, wiToolsPath uses malloc inside.

RafaelLeeImg:
Pretty good, this PR fixes the double double free error, works well on my Debian.
I confimed the base 491526f1e069328feacc54db7c7d82b62a2cb89c has double free problem.
I checked the modified lines, it solved the problem.
All dynamically allocated resources are freed in lib/osdep/linux.c, line 2279
```C
static void do_free(struct wif * wi)
```

jbenden:
I have added missing pieces from this PR to current master.

Thanks!
-Joe

PR #2095, opened 2019-11-19
fix src/Makefile.inc to get --enable-static --disable-shared and opti…

dmfreemon:
…onally --with-static-simd=xxx to build successfully


jbenden:
In the future, do not merge to your PR branch; but rather, use rebase.

PR #2102, opened 2019-12-03
bugfix: fix memory references in dump_print() after main thread has...

dmfreemon:
free() the memory and also prevent corruption if another thread changes the elapsed_time string _during_ the sprintf.


dmfreemon:
I have no idea why this PR is failing in Travis

jbenden:
Hi!

It's because the source code is not properly formatted with clang-format 3.8.

-Joe

dmfreemon:
I appreciate the timely reply, but I've been through the Travis output several times, and I don't see any information about a formatting error, what formatting rule was violated, or which source code line number has the problem.  Nothing.  So there is nothing I can do about this.  In reviewing my changes associated with this PR, I'm comfortable with what I am presenting to you guys.  So at this point, I'll just leave this PR with you for your consideration, to either commit, defer, reject, or commit with whatever formatting (or other) changes you want.

aircrack-ng:
Just click on the Travis Build, which will lead to 4 items, click on the one that is red, and expand "Checking source code style...", then go to the end of it.

It's marked failed because codestyle but the build is succesfull.

dmfreemon:
There may be a problem with Travis.  For this PR, the code (I think) conforms to the formatting standards.  In particular, there are some if statements with the opening braces on the same line.  See:
https://github.com/aircrack-ng/aircrack-ng/pull/2102/files

But Travis, when it performs the code formatting check, gets it backwards, and thinks I'm putting the opening brace on a new line. See:
https://travis-ci.org/aircrack-ng/aircrack-ng/jobs/654115815?utm_medium=notification&utm_source=github_status

PR #2105, opened 2019-12-07
Improve some string handling that could lead to SEGVs; also eliminate the FORTIFY_SOURCE warnings

dmfreemon:
Improve some string handling that could lead to SEGVs; also eliminate the FORTIFY_SOURCE warnings


dmfreemon:
Here is another example of two things:
* "make check" tests that pass on my machine but fail on the build machines
* build failures cause by code formatting issues, but no indication in the Travis output which lines of code violated the formatting rules, nor even _which rules_ were violated.  How can I get better output from the formatting checker?

aircrack-ng:
Click on the one that fails on Travis, then click on the yellow "Checking source code style...". It will open it. Scroll down, and you'll see a diff showing what's wrong.

dmfreemon:
> Click on the one that fails on Travis, then click on the yellow "Checking source code style...". It will open it. Scroll down, and you'll see a diff showing what's wrong.

I do get that far, and I do see the diff.  But there are a lot of lines on the diff, with no indication of what lines are being flags as errors.  There is also no indication of what rule is being violated, or how many rules are being violated.  I took a quick look, at did not see an obvious way for me to run these code formatting checks myself (in my local workspace).  Is there a way to do that?  What is the normal process for fixing these?  I am guessing it's not supposed to be: guess, commit, push, guess, commit, push...

aircrack-ng:
What's in red is the problem, what's in green corrected.

Yes, use clang 3.6 (or 3.8) and use the provided [.clang-format](https://github.com/aircrack-ng/aircrack-ng/blob/master/.clang-format) file. You can also use [.editorconfig](https://github.com/aircrack-ng/aircrack-ng/blob/master/.editorconfig)

dmfreemon:
Ok, well, I install the clang tools and run the code formatter with the rules provided in the aircrack repository (.clang-format file), and the build _still_ fails, due to formatting problems.  See this PR, commit dfca6bb.

aircrack-ng:
It's not just any clang-format, it's a specific version, as they change from one version to another. As mentioned earlier it is clang-3.8

dmfreemon:
I am on Fedora 31, which provides clang versions 7.0.1, 8.0.0, and 9.0.0 in the dnf repositories.

PR #2112, opened 2019-12-28
is aircrack combatible edup ed 200 wifi adapter? 

yusufsayar:
is aircrack combatible edup ed 200 network wifi adapter on rpi3? thanks


aircrack-ng:
What you are asking is for the forum, not for github, as it is a support request.

May I also suggest to read the documentation, we have plenty in the wiki that helps determining the chipset, and thus if the card will work.

Issue #2096, opened 2019-11-19
Use of dynamic shared objects


dmfreemon:
I have been investigating some sporadic seg faults in airodump and found some things in the string manipulation (and at the same time eliminating the FORTIFY_SOURCE warnings).  In testing the changes, I've hit issues with some symbols during linking related to the use of dynamic shared objects.  In thinking about that, I started to wonder if it's even necessary to use dynamic shared objects at all (specifically, I'm referring to libaircrack-ce-wpa[.*] and libaircrack-osdep.  Everything else is statically linked.)

I did a quick comparison of the size differences in the executables and found:
```
                                   size when using                         size when using
                                dynamic (shared objects)		   static linking
airbase-ng				139848			airbase-ng	184032
aircrack-ng				240120			aircrack-ng	251392
airdecap-ng				103536			airdecap-ng	103536
airdecloak-ng				52144			airdecloak-ng	52240
aireplay-ng				156432			aireplay-ng	204560
airodump-ng				191744			airodump-ng	235040
airolib-ng				83352			airolib-ng	83352
airserv-ng				38592			airserv-ng	83200
airtun-ng				101456			airtun-ng	150088
ivstools				69664			ivstools	65568
libaircrack-ce-wpa-1.5.2.so		86432				
libaircrack-ce-wpa-x86-avx-1.5.2.so	86552				
libaircrack-ce-wpa-x86-avx2-1.5.2.so	86552				
libaircrack-ce-wpa-x86-sse2-1.5.2.so	90648				
libaircrack-osdep-1.5.2.so		75864				
makeivs-ng				37784			makeivs-ng	37784
packetforge-ng				92640			packetforge-ng	145360
wpaclean				27848			wpaclean	77496
```
The differences in executable size is minimal.  It's also unclear if there is much practical opportunity for memory savings by sharing the same library code across many processes. 
Discontinuing support of building these 5 libraries as dynamic shared objects would certainly simplify the makefiles and make it easier to diagnose build problems.
Are there other motivations for supporting the use of shared objects for these 5 libraries?

PR #2127, opened 2020-02-22
Remove support for building aircrack libraries as dynamic shared objects

dmfreemon:
This PR removes the option to build the aircrack libraries as dynamic shared objects.

The motivation is that supporting building DSO's makes the build system more complicated, without offering any benefits (in this case).

Although more simplifications are possible, this set of changes works and is good enough to commit.

For more background and details, see issue #2096:
#2096


jbenden:
This is a significant change which should have been discussed prior to performing programming work.

Tossing aside the failed CI builds; this PR cannot be accepted.

The purpose of the multiple dynamic libraries is to support "single compilation" and "multiple machine use". This is common with Linux distributions, as they build the packages and many different people on different machines use the software.

The area in which the multiple dynamic libraries exists is to handle cryptographic calculations, which are the perfect candidate for the added complexities involved. This is typical and the norm - see openssl, etc. They all have modules split on architecture and features of certain architectures, such as AVX, SSE2, ASIMD, NEON, etc....

PR #2140, opened 2020-04-25
Update Travis CI

CGarces:
- OSX custom code replaced with homebrew addon.
- Removed packages that already exist on current linux and OSX image
- Enabled ccache.

I'm not sure if Travis CI is really used, I have several questions before set this PR as ready to merge.

1. Code looks outdated, maybe the original code was designed for pangolin, as I see at:
https://github.com/aircrack-ng/aircrack-ng/blob/master/build/travis-formatting.sh#L17
There is no `distro` tag set to is using Xenial. ¿can I update to Bionic?

2. OSX image is using xcode9.4 that has one bug on brew addon. The bug is fixed on xcode11. ¿can I update to xcode11? I'm not familiar with OSX.

 3. I fact in this PR the OSX build **not install** cmocka and libpcap, and looks like is working fine. ¿this dependencies are needed?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,18 @@ os:
     - linux
     - osx
 
+#need due bug on homebrew see 
+#https://changelog.travis-ci.com/xcode-11-3-1-xcode-11-2-1-xcode-11-1-and-xcode11-images-updated-142286
+osx_image: xcode11.4
+
 language: c
 
-sudo: true
-
 before_install:
-    - if [ "$TRAVIS_OS_NAME" == "osx" ]; then (brew update && brew install cmocka openssl libgcrypt sqlite3 libpcap pcre || echo -e "\e[1;31m Skipping Homebrew... \e[0m"); fi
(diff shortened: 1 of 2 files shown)


CGarces:
Another question.
`cpp-coveralls` is installed on linux targets, but ¿where is used?

Looks like was removed on version 1.6
https://github.com/aircrack-ng/aircrack-ng/blob/e708c21e7a0f88c65d1ea5f411b1db78a2e314c0/ChangeLog#L44

aircrack-ng:
That is correct, it's not used. It was used at some point but removed since we were not using it.

CGarces:
> That is correct, it's not used. It was used at some point but removed since we were not using it.
 
It's not used cpp-coveralls or the entite .travis.yml?

CGarces:
Removed custom code to download clang-format-3.8 
I'm not sure if the OSX dependencies are really needed, script build fine without it.
Ready for review in any case.

jbenden:
Hi,

I've been bitten in the past by NOT specifying all dependencies; so I'd prefer that they all be included (as they are currently).

As for the rest, it looks great!

Thanks,
-Joe

CGarces:
I have restored the dependencies, as requested.

jbenden:
Thanks a bunch!

PR #2146, opened 2020-05-22
airbase: add -R (bssid randomization)

beansrus:
Hello, this is a small little patch to add a '-R' argument to airbase to be able to pick a random bssid.

The OUI of the generated mac has the locally administered and unicast bits set.

I also changed adhoc to use the same function since it was doing the same thing (minus enforceing  the local admin bit).

This is my first pull-request to the project, please let me know if I'm missing anything.


beansrus:
Hmm, doesn't look like these check failures are due to my changes

jbenden:
@Mister-X- Is this acceptable for merge?

PR #2152, opened 2020-06-07
Colorize airodump new bug fix

scorpius:
This **finally** fixes the previous terminal issue: _"existing clients not linked to AP in the top section become hidden when hitting 'o'. This could be..."_

Now those clients are displayed in red, and each other client matches the color of the AP it is associated to.


SNiTEBoBy:
Pull merged

Issue #2154, opened 2020-06-23
airodump-ng memory double-free


RafaelLeeImg:
# Issue type

- Defect - Crash
# System information

- OS: Debian testing
- CPU: Intel(R) Core(TM) i7-6700K

# Aircrack-ng version

- Version: master
- Commit Revision hash: 85754a8cb5fe8b87a24d1e88d961d81741019a5d

# Defect
## How to reproduce the issue
```bash
sudo ./airodump-ng wlan0mon
```
When program exits, it complains:
```bash
free(): double free detected in tcache 2
```

When compile and link with AddressSanitizer, add "-fsanitize=address" in CFLAGS, CXXFLAGS, LDFLAGS in Makefile, it shows that
```bash
==665629==ERROR: AddressSanitizer: attempting double-free on 0x6120000001c0 in thread T0:
    #0 0x7f20b9d05277 in __interceptor_free (/usr/lib/x86_64-linux-gnu/libasan.so.5+0x107277)
    #1 0x7f20b9bd3da9 in do_free lib/osdep/linux.c:2288
    #2 0x7f20b9bd3f91 in linux_close lib/osdep/linux.c:2319
    #3 0x7f20b9bcbfba in wi_close lib/osdep/osdep.c:97
    #4 0x557f712a6f2a in main src/airodump-ng/airodump-ng.c:7339
    #5 0x7f20b969de0a in __libc_start_main ../csu/libc-start.c:308
    #6 0x557f712ac2d9 in _start (/dev/shm/d/aircrack-ng/.libs/airodump-ng+0x2b2d9)

0x6120000001c0 is located 0 bytes inside of 263-byte region [0x6120000001c0,0x6120000002c7)
freed by thread T0 here:
    #0 0x7f20b9d05277 in __interceptor_free (/usr/lib/x86_64-linux-gnu/libasan.so.5+0x107277)
    #1 0x7f20b9bda40d in do_linux_open lib/osdep/linux.c:2262

previously allocated by thread T0 here:
    #0 0x7f20b9d0580e in calloc (/usr/lib/x86_64-linux-gnu/libasan.so.5+0x10780e)
    #1 0x7f20b9bd4121 in searchInside lib/osdep/linux.c:211

SUMMARY: AddressSanitizer: double-free (/usr/lib/x86_64-linux-gnu/libasan.so.5+0x107277) in __interceptor_free
==665629==ABORTING
```

[...]

PR #2155, opened 2020-06-23
Fix a double-free problem of dev->iwpriv

RafaelLeeImg:
Details are described in:
#2154

### DIFF
--- a/lib/osdep/linux.c
+++ b/lib/osdep/linux.c
@@ -2260,6 +2260,7 @@ static int do_linux_open(struct wif * wi, char * iface)
 
 	if (iface_malloced) free(iface);
 	if (iwpriv) free(iwpriv);
+	dev->iwpriv = NULL;
 	if (r_file)
 	{
 		free(r_file);
@@ -2276,6 +2277,7 @@ close_in:
 	if (acpi) fclose(acpi);
 	if (iface_malloced) free(iface);
 	if (iwpriv) free(iwpriv);
+	dev->iwpriv = NULL;
 	return 1;
 }
 


dmfreemon:
See #2094 (from November 2019)

RafaelLeeImg:
Thanks.
I saw that, actually the pointer not referring to a static array.
The free of iwpriv is required. It is not pointing to the element of the static array.

```C
// aircrack-ng/lib/osdep/linux.c, line 1869
iwpriv = wiToolsPath("iwpriv");


// aircrack-ng/lib/osdep/linux.c, line 255
static char * wiToolsPath(const char * tool)
{
    static const char * paths[] = {"/sbin",
                                   ...};
    for (i = 0; i < nbelems; i++)
    {
        path = searchInside(paths[i], tool);
        if (path != NULL) return path;
    }
}


// aircrack-ng/lib/osdep/linux.c, line 194
static char * searchInside(const char * dir, const char * filename)
{
    ...
    curfile = (char *) calloc(1, lentot);

    while (searching)
    {
        if (match)
        {
            // not freed here
            return curfile;
        }
    }

    (void) closedir(dp);
    // free only if curfile not return
    free(curfile);
    ...
}
```

dmfreemon:
> I saw that, actually the pointer not referring to a static array.

Looking at PR #2094 , there is a comment at the top of the wiToolsPath function in linux.c that says:  "Note: The returned string is a new allocation from the heap."

I believe you may be referring to a comment from someone else.

> The free of iwpriv is required.

Are you sure that do_free() via linux_close..() doesn't take care of that free (after PR #2094 is applied)?

RafaelLeeImg:
The code I wrote above is only pseudo code describing the work flow, not from any file from project.
I checked #2094 , the way @dmfreemon deals with device pointer is consistent.
I think if that PR merges, the problem will be solved in a systematic method. Then my PR should be invalid and not required.
I will spend more time checking #2094

RafaelLeeImg:
Same to #2094

PR #2177, opened 2020-08-25
airbase: fix beacon interval timing

beansrus:
Hello.  This pull request tries to fix up some of the beacon timing in airbase that I came across.  Please let me know if there is anything I need to fixup to be accepted.

---

The current implementation had a few problems.

* the beacons timing would not scale correctly for multiple ssids.  If
  the time was 100 ms, and 5 ssids were beaconing, each ssid would beacon
  every 500ms instead of every 100ms.
* The beacon timing would drift after some time.
* The gettimeofday() is affected by discontinuous jumps in the system
  time (NTP adjustments and such).
* the rtc code seemed to over complicate the timing code.  Beacon
  frames are usually 100ms.  A timer that takes into account small
  amounts of jitter should be sufficient for beacon timing.

Tested with 50 ssids, on a ARMv7 @ 400mhz and didn't have any problem
keeping up.  I saw jitter somewhere between 400us-1.5ms

### DIFF
--- a/src/airbase-ng/airbase-ng.c
+++ b/src/airbase-ng/airbase-ng.c
@@ -140,6 +140,14 @@
 	"\xf2\x01\x00\x50\xf2\x02\x00\x50\xf2\x03\x00\x50\xf2\x04\x00\x50"         \
 	"\xf2\x05\x02\x00\x00\x50\xf2\x01\x00\x50\xf2\x02"
 
+#define USEC_TO_MSEC(x) ((x) / 1000UL)
+#define NSEC_TO_USEC(x) ((x) / 1000UL)
+#define NSEC_TO_MSEC(x) ((x) / 1000000UL)
+#define SEC_TO_USEC(x) ((x) *1000000UL)
+#define SEC_TO_MSEC(x) ((x) *1000UL)
+#define SEC_TO_MSEC(x) ((x) *1000UL)
+#define MSEC_TO_USEC(x) ((x) *1000UL)
+
 static const char usage[]
 	= "\n"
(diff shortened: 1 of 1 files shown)


jbenden:
Hi,

Would you mind rebasing on top of the lastest `master` branch?

Thanks!
-Joe

beansrus:
Hello! rebased on master.  Quickly gave it a test run and it still seems to work.

jbenden:
I am concerned that big-endian machines will fail. An example is the timestamp; this value is internally CPU native but has been encoded in fixed form dependent on native CPU endianess.

beansrus:
Do you mean the part like  (```beacon[24 +i] = (uint8_t)((timestamp >> (i *8)) & 0xFF)```) ?  This is existing code.  You can see the same usage pattern a couple of times in the file.  

Also, it seems to be ok, as its updating the beacon 1 byte at a time so byte ordering shouldn't come into play.

jbenden:
@Mister-X- Is this PR ready for merge?

Mister-X-:
By the way, monotonic clock is still affected by jumps of time, like `gettimeofday()`.

If I'm understanding correctly we'd have to use the TAI clock to be unaffected, however, that's Linux only.

Mister-X-:
Could you also check and fix codestyle? We have a `.clang-format` file at the root that can be used with clang-format 3.8.

beansrus:
Hmm, I updated the comment about the NTP, perhaps I must have been thinking about monotonic_raw.  Also, I don't see any changes with clang 3.8, although I do with clang-format 14 (was already installed in my system).  Perhaps I missed something

Issue #1795, opened 2018-03-10
besside-ng hops only on channels 1-11 instead of all of the available ones


aircrack-ng:
**Reported by strasharo on 3 Apr 2017 17:05 UTC**

It appears that by default in channel hopping mode besside-ng is hopping on channels 1-11 instead of 1-14 even if all of the channels are unlocked on the adapter:
```

pi@zero3:~ $ iw phy phy1 info | grep MHz
                        * 2412 MHz #1 (33.0 dBm)
                        * 2417 MHz #2 (33.0 dBm)
                        * 2422 MHz #3 (33.0 dBm)
                        * 2427 MHz #4 (33.0 dBm)
                        * 2432 MHz #5 (33.0 dBm)
                        * 2437 MHz #6 (33.0 dBm)
                        * 2442 MHz #7 (33.0 dBm)
                        * 2447 MHz #8 (33.0 dBm)
                        * 2452 MHz #9 (33.0 dBm)
                        * 2457 MHz #10 (33.0 dBm)
                        * 2462 MHz #11 (33.0 dBm)
                        * 2467 MHz #12 (33.0 dBm)
                        * 2472 MHz #13 (33.0 dBm)
                        * 2484 MHz #14 (33.0 dBm)
                 * short GI for 40 MHz
pi@zero3:~ $
```

If I specify any of the missing channels manually by using the -c parameter it's able to switch to those channels.
Is it possible to have it recognize the available channels to the adapter (like kismet does) instead of defaulting to 1-11?

Thank you.

aircrack-ng:
**Comment by strasharo on 2 May 2017 10:54 UTC**

https://github.com/aircrack-ng/aircrack-ng/pull/111 fixes this, just tested it.

Issue #1796, opened 2018-03-10
Besside-ng: Added automatic channel detection of 2GHz and 5GHz channels


aircrack-ng:
**Reported by moonrk on 30 Apr 2017 20:11 UTC**

The channels in besside-ng were hardcoded from 1 to 11 unless using the -c option to lock to a specific channel.

I've made the necessary changes to detect the channels that the wifi adapter supports automatically, including 5GHz channels. The scanning process will thus be able to scan channels 1-14, 36-64, 100-144, 149-161 and 165 if they are supported.

This also fixes issue #106.

Tested on Alfa AWUS036NHA and Alfa AWUS036AC wifi adapters.

aircrack-ng:
**Pull Request by moonrk on 30 Apr 2017 20:11 UTC**

Source Location: moonrk:master
Target Location: master

aircrack-ng:
**Comment by strasharo on 2 May 2017 10:56 UTC**

Awesome, this works, just tested it on AWUS036H. Thanks a lot @moonrk .
@Mister-X- , can we have this merged, please? :)

aircrack-ng:
**Comment by Mister X  on 2 May 2017 19:16 UTC**

It's nice to have it in bessides-ng but that needs to be global, in osdep, so airodump-ng can query it. And split it in 2 parts, 2.4GHz and 5GHz.

Other tools also need to be able to query which channels are available (to validate them), so a third function to check that in osdep would be useful.

SymbianSyMoh:
I did this solution to temporary over come the issue :D 

![screen shot 2019-01-04 at 8 46 36 pm](https://user-images.githubusercontent.com/3141848/50719476-fd5b4500-1061-11e9-8ef9-bcd3428a0a59.png)

PR #2193, opened 2020-10-28
Support ALL Channels for 5GHz - ported from @moonrk

pect0ral:
Ported changes proposed by @moonrk to support and detect all channels in Auto-Channel mode against current latest code. 

Tested working with AWUS036ACH and working great!


pect0ral:
This should fix issue #1795 and #1796 -- I've updated the code from @moonrk to fit the latest code base.

This PR specifically targets besside-ng, and nothing else. Builds clean.

PR #2196, opened 2020-11-14
Fixed issue causing "No such BSSID available." on higher 5GHz-channels

matssoderhall:
If the normal channel-tag does not exist we look for channel-info in the HT-info.
Also, made some changes that i hope will make it easier to understand what is happening in the code.


jbenden:
Thanks for your submission! After some successful local testing, I've committed a modified version of your change; fixing some spelling mistakes and single-character test logic.

Thanks again,
-Joe

Issue #1707, opened 2018-03-10
airodump-ng provides "faulty" CSV dump files due to BSSIDs that contain commas


aircrack-ng:
**Reported by danielrodriguezac on 5 Dec 2016 17:43 UTC**

I have encountered a situation where using the .csv dump file is very useful, however i have found that if the BSSID of the station contains a comma the CSV becomes unwieldy, since i am processing as CSV it would be quite useful if the values where escaped or encased within double quotes (") as recommended by the most commonly accepted "spec". Why this is not the case evades me, trying to save space maybe?. I saw another ticket with a patch to output JSON, might be i can use that in the future but for now is there any way to tell airodump-ng to add the quotes?

Thanks for the help and pardon the broken english.

aircrack-ng:
**Comment by misterx on 5 Dec 2016 17:43 UTC**

Could you upload the offending CSV?

If you are talking about the probes, how should we separate the different SSID? What if the same separator is used in a SSID?

aircrack-ng:
**Attachment by danielrodriguezac on 6 Dec 2016 14:02 UTC**

Uploaded file [cap-01.csv](https://github.com/aircrack-ng/trac-attachments/blob/master/1707/20161206-140246_047-cap-01.csv)

Offending CSV

aircrack-ng:
**Comment by anonymous on 5 Dec 2016 17:43 UTC**

Of course,should be readable now, kinda new at this so please excuse any errors. From what i have read the important part of meeting the "spec" is to encase the field in double quotes, even if there are double quotes inside the parser should be able to deal with that. On the file look at the last AP line, daniel,AP.

aircrack-ng:
**Comment by misterx on 5 Dec 2016 17:43 UTC**

You are probably using an older version. I tried with a comma separated SSID and it escapes it just fine.

PR #2206, opened 2020-12-25
airgraph-ng, airdrop-ng: add ability to parse ESSIDs containing commas from .csv files

mehw:
Hello,

Here I propose some amendments to allow `airgraph-ng` and `airdrop-ng` to parse ESSIDs containing commas from .csv files.

### Summary of bugs found
- `airgraph-ng ... -g CAPR`: entries with an ESSID containing commas in its name being discarded.
- `airgraph-ng ... -g CPG`: _Probed ESSIDs_ are truncated when their name contains commas.

See bottom about the [dump-01.txt](https://github.com/aircrack-ng/aircrack-ng/files/5742324/dump-01.txt) test file (aka `dump-01.csv`).

### Description
Calling `airgraph-ng ... -g CAPR` I could see **entries with an ESSID containing commas in its name being discarded**. This is due to requirements/processing in `airgraphviz/libDumpParse.py` (which are in `airdrop/libDumpParse.py` too):
1. The entry with the offending ESSID is read from the _csv-comma-separated_ file;
2. the read entry is then comma-split (`string_list = entry.split(',')`);
3. the number of fields are expected to be 15 or 11 (`if len(string_list) == 15` or `if len(string_list) == 11`);
4. an ESSID containing commas may exceed the number of expected fields or even give corrupted values (think of an entry of 11 fields, where the ESSID contains enough commas to produce 15 fields when it is comma-split, the resulting 15-fields holds corrupted values).

See also #1707 as reference (the author commented about a [cap-01.csv](https://github.com/aircrack-ng/trac-attachments/blob/master/1707/20161206-140246_047-cap-01.csv) comma-related problem...).

There's also a problem calling `airgraph-ng ... -g CPG`. **The _Probed ESSIDs_ are truncated when their name contains commas**. This is due to `string_list = entry.split(',')`. Since probed ESSID names are comma separated, you could easily see how a name containing commas is split (like **this is, my ap** is split into **this is** and **my ap**).

### Solution
I made some changes to `{airgraphviz,airdrop}/libDumpParse.py` and `airodump-ng/dump_write.c` to address the problems described above. In the former there's some regex used to determine the ESSID position in the entry, and in the latter each ESSID probe wrote in the .csv file (aka _Probed ESSIDs_) is preceded by its length. The ESSID length is used to grab the whole ESSID name without exception (there's also a fallback on `entry.split(',')` fields for .csv files that are in the old or wrong format):
- **this is, my ap** ESSID name should not be discarded when producing a CAPR.
- **name with , comma** ESSID probe should not be truncated when producing a CPG.
- AP entries may end with the previous format "**lan_ip, essid**" (one space after the comma).
- AP entries may end with the current format "**essid_length, essid, key**" (one space after the comma).
- _Probed ESSIDs_ should follow the **NEW FORMAT** "**essid_length,essid[,...]**" (no spaces after the comma).

### Tests
The file [dump-01.txt](https://github.com/aircrack-ng/aircrack-ng/files/5742324/dump-01.txt) is conceived to test the amendments.

PS: I renamed `dump-01.csv` to `dump-01.txt`, because .csv files cannot currently be attached to posts.

```
$ airgraph-ng -i dump-01.txt -o dump-01-cpg.png -g CPG
$ airgraph-ng -i dump-01.txt -o dump-01-capr.png -g CAPR
```

Let me know what you think.

Thanks!

### DIFF
--- a/scripts/airdrop-ng/airdrop/libDumpParse.py
+++ b/scripts/airdrop-ng/airdrop/libDumpParse.py
@@ -122,16 +122,35 @@ class airDumpParse:
 		dict = {}
 		for entry in devices:
 			client = {}
+			# NOTE: It is expected a rstripped entry.
 			# WARNING: Splitting the entry on every comma means to split the ESSID too if it contains a comma, resulting in more items than expected.
 			string_list = entry.split(',')
 			if len(string_list) >= 7:
+				 probe_list = []
+				 p = re.match(r'^([^,]+,){6}(.*)$',entry).group(2)
+				 while p:
+					 # expect essid_length,essid_string[,...]
+					 essid_length = re.match(r'^([0-9]{1,2}),',p)
+					 if not essid_length:
+						 # this may happen if the entry is malformed or incompatible
+						 probe_list = string_list[6:][0:]
+						 break
+					 essid_length = essid_length.group(1)
+					 l = len(essid_length)
+					 n = int(essid_length)
+					 if n > 32:
+						# this may happen if the entry is malformed or incompatible
+						probe_list = string_list[6:][0:]
+						break
+					 probe_list.append(p[l+1:l+1+n])
+					 p = p[l+2+n:]
 				client = {"station":string_list[0].replace(' ',''),
 					"fts":string_list[1],
 					"lts":string_list[2],
 					"power":string_list[3],
 					"packets":string_list[4],
 					"bssid":string_list[5].replace(' ',''),
-					"probe":string_list[6:][0:]} # ESSIDs cannot be split faithfully if any contains a comma (see the WARNING above)
+					"probe":probe_list} # ESSIDs cannot be split faithfully if any contains a comma (see the WARNING above)
 			if len(client) != 0:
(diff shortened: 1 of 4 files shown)


aircrack-ng:
Actually, the issue starts in airodump-ng, who should escape the SSID if it contains a comma, according to the [Wikipedia page about CSV](https://en.wikipedia.org/wiki/Comma-separated_values#Example)

Basically, the whole field need to be enclosed with double quotes. So an ESSID of `test,` should be `"test,"`. And if there is a double quote in there, it should be escaped as well with a double quote, `test,"` should become `"test,"""`. Same goes if there is a line break.

However, we might have another problem as there apparently shouldn't be a space character between the comma and the start of the field. Which means, we should check other wireless tools making use of this CSV and see how they handle it, and discuss the appropriate way to fix this.

aircrack-ng:
Let's close this as it needs to be fixed differently.

However, that is a well written/structured (Summary of bugs found, Description, Solution, Tests) PR.

PR #2249, opened 2021-07-18
Update INSTALLING

ghost:
plugin info


aircrack-ng:
While it probably should (and that would be a PR by itself), airoscript isn't installed with with that flag, so this PR will be closed.

Dependencies are listed in the README of the individual directories (`airdrop-ng`, `airgraph-ng`, and `versuck-ng`). Adding that information (not each of the dependencies but a sentence similar to the previous sentence) in the README.md is acceptable.

Issue #2205, opened 2020-12-18
Add support for WiFi-6E (6Ghz) Channels


wlanwizard:
The Intel AX210 WiFi adapter supports the new 6Ghz band but airodump-ng does not. airmon-ng will successfully put the card into monitor mode however airodump-ng fails when specifying a 6Ghz channel such as 233. Using channel 173 puts the card to channel 1 (2.4Ghz)

super@super-desktop:~$ sudo airodump-ng -c 233 wlp2s0mon

  Airodump-ng 1.6  - (C) 2006-2020 Thomas d'Otreppe
  https://www.aircrack-ng.org

  usage: airodump-ng <options> <interface>[,<interface>,...]
...

Tested with Ubuntu 20.04 running 5.10.1-051001-generic kernel with ver 1.187 linux-firmware

PHY	Interface	Driver		Chipset
phy0	wlp2s0		iwlwifi		Intel Corporation Device 2725 (rev 1a)

		(mac80211 monitor mode vif enabled for [phy0]wlp2s0 on [phy0]wlp2s0mon)
		(mac80211 station mode vif disabled for `[phy0]wlp2s0)`

iwconfig:
wlp2s0mon  IEEE 802.11  Mode:Monitor  Frequency:2.457 GHz  Tx-Power=-2147483648 dBm   
          Retry short limit:7   RTS thr:off   Fragment thr:off
          Power Management:on

Supported band 2 channel list iw list:
  Frequencies: 

            * 5180 MHz [36] (22.0 dBm) 

            * 5200 MHz [40] (22.0 dBm) 

            * 5220 MHz [44] (22.0 dBm) 

            * 5240 MHz [48] (22.0 dBm) 

            * 5260 MHz [52] (22.0 dBm) 

            * 5280 MHz [56] (22.0 dBm) 

            * 5300 MHz [60] (22.0 dBm) 

            * 5320 MHz [64] (22.0 dBm) 

            * 5340 MHz [68] (22.0 dBm) 

            * 5360 MHz [72] (22.0 dBm) 

            * 5380 MHz [76] (22.0 dBm) 

            * 5400 MHz [80] (22.0 dBm) 

            * 5420 MHz [84] (22.0 dBm) 

            * 5440 MHz [88] (22.0 dBm) 

            * 5460 MHz [92] (22.0 dBm) 

            * 5480 MHz [96] (22.0 dBm) 

            * 5500 MHz [100] (22.0 dBm) 

[...]

PR #2273, opened 2021-11-03
added Wi-Fi 6E 6GHz channels

prash42135:
Can be used to capture 6GHz using Intel AX210 as:
sudo airmon-ng check kill
sudo airmon-ng start wlan1
sudo airodump-ng wlan1mon -w 6E_test -C 6115

is tested and working in:
Ubuntu 20.04.3 LTS
Linux Kernel 5.15.0-051500-generic

### DIFF
--- a/src/airodump-ng/airodump-ng.c
+++ b/src/airodump-ng/airodump-ng.c
@@ -5701,9 +5701,9 @@ static int detect_frequencies(struct wif * wi)
 		}
 	}
 
-	// again for 5GHz channels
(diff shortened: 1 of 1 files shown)


ZeroChaos-:
https://github.com/aircrack-ng/aircrack-ng/issues/2205

ZeroChaos-:
This seems fine to me, but I'll wait for @Mister-X- to approve it, probably.

ZeroChaos-:
This only extends "-C 0" support but it is the right change so, accepted.  Thanks!

PR #2277, opened 2021-11-08
airodump-ng: fix hang on quit while GPSD is down or no device messages appear

### LINKED ISSUE #2276: Airodump-ng: Hangs when quitting while GPSD is down or no device messages appear
# Defect Report

## Issue type

- [x] Defect - Unexpected behavior (obvious or confirmed in the forum)

## Aircrack-ng version

- Version: master
- Commit Revision hash: https://github.com/aircrack-ng/aircrack-ng/commit/6ac65c89c8bf672494ce64055717676f0ab2ea38

## Defect

When exiting airodump-ng by either keypress 2x 'q', CTRL-C or SIGTERM signal, when GPSD is activated and connected using --gpsd option and GPSD stopped providing (arbitrary) messages, the program doesn't exit. Only a SIGKILL signal will end it.

### How to reproduce the issue

1. have GPSD running with connected GPS device
2. run `airodump-ng --gpsd <wlandev>` 
3. prevent GPS device from providing GPSD with messages, e.g. remove USB or Bluetooth connection
4. try to exit airodump-ng by 'qq', CTRL-C or SIGTERM

n0w1re:
When airodump-ng is exiting, the main thread waits for the gps thread to finnish when the exit flag is set.
If GPSD is not delivering any message to the connected socket, that thread hangs in a blocking `recv` call until some data is received, possibly indefinitely so the program doesn't exit.

Fix:
- a `select` waits for available data on the socket before `recv` is called
- when the `select` timeouts every 500ms the global exit flag is checked
- both code paths for newer JSON based and old non-JSON based GPSD are fixed

This PR fixes #2276

### DIFF
--- a/src/airodump-ng/airodump-ng.c
+++ b/src/airodump-ng/airodump-ng.c
@@ -4388,6 +4388,10 @@ static inline ssize_t
 read_line(int sock, char * buffer, size_t pos, size_t size)
 {
 	ssize_t status = 1;
+	fd_set read_fd;
+	int fdnum;
+	struct timeval timeout;
+
 	if (size < 1 || pos >= size || buffer == NULL || sock < 0)
 	{
 		return (-1);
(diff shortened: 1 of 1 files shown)


jbenden:
Hi, could you please rebase from master and push PR again? Changes were made to fix macOS and the PVS-Studio analyzer. Thanks!

timg11:
I just installed Airodump-ng 1.7 rev d5cb1507 in a new Ubuntu 22.10 system, and I'm seeing this issue.  Has the fix-quitgpsdhang branch from 2 years ago not yet been merged? Or is the Ubuntu repository installed with `sudo apt install aircrack-ng` out of date?

Mister-X-:
Considering the PR is still open, that means it hasn't been merged.

This needs some testing as it seem to affect more than just GPSd:
- Testing current, and reproducing this bug
- Testing current, without GPS, but with airserv-ng, with 1. capture only, 2. injection only, 3. capture and injection, and 4. multiple users
- Same as above, with GPSd as well.
- Repeat all of the above testing with this patch applied to see it fixes GPSd, and that it doesn't break other things.

If you have time to do that testing, that would greatly help us.

PR #2286, opened 2021-11-25
airodump-ng: look for oui.txt in /usr/share/hwdata

floppym:
This location is used by the hwdata package on various distros.

https://github.com/vcrhonek/hwdata


ZeroChaos-:
ACK, thanks!

PR #2317, opened 2022-06-27
airmon-ng.linux: Unquote newlines when parsing sysfs

tyll:
When creating iface_list from sysfs, also call printf to unquote `\n`
characters and sort the interface list.

### DIFF
--- a/scripts/airmon-ng.linux
+++ b/scripts/airmon-ng.linux
@@ -1253,6 +1253,7 @@ scanProcesses() {
 
 listInterfaces() {
   unset iface_list
+  sort_command="cat"
   for iface in $(ls -1 /sys/class/net)
   do
     if [ -f "/sys/class/net/${iface}/uevent" ]; then
@@ -1263,14 +1264,15 @@ listInterfaces() {
     fi
   done
   if [ -x "$(command -v iwconfig 2>&1)" ] && [ -x "$(command -v sort 2>&1)" ]; then
+    sort_command="sort -bu"
     for iface in $(iwconfig 2> /dev/null | sed -e 's/^\(\w*\)\s.*/\1/' -e '/^$/d'); do
       iface_list="${iface_list}\n ${iface}"
     done
-    # This breaks if you fix SC2059 and I am not sure why
-    # This should probably be space separated instead of newline
-    # shellcheck disable=SC2059
-    iface_list="$(printf "${iface_list}" | sort -bu)"
   fi
+  # This breaks if you fix SC2059 and I am not sure why
+  # This should probably be space separated instead of newline
+  # shellcheck disable=SC2059
+  iface_list="$(printf "${iface_list}" | ${sort_command})"
 }
 
 getPhy() {


ZeroChaos-:
I will test this, the diff isn't helpful on my phone.

tyll:
Thank you for the quick feedback.

I forgot to update the indentation of the comment, the diff might be easier to read now.
https://github.com/aircrack-ng/aircrack-ng/blob/master/scripts/airmon-ng.linux#L1254 is the original function, there you can see that `printf "${iface_list}"` is only called when iwconfig was parsed in https://github.com/aircrack-ng/aircrack-ng/blob/master/scripts/airmon-ng.linux#L1265 but not when sysfs is parsed. But when sysfs is parsed, newlines are also added in https://github.com/aircrack-ng/aircrack-ng/blob/master/scripts/airmon-ng.linux#L1261

So without this patch, `iface_list` contains strings that literally contain a backslash and an `n` instead of space-separated interfaces.

tyll:
@ZeroChaos- is there anything I can do to move this forward?

jbenden:
@tyll Line 1265 tests for the `sort` command; which was used in that block. This patch moves its use outside of that block. It needs fixed somehow. :)

ZeroChaos-:
I finally had time to review this, and the whole thing makes no sense.  It doesn't make sense to run through sort -u if we didn't get devices from two different sources which may provide duplicates.  I agree this needs to be fixed, but not in the way which it's being fixed.

Is the simple test case to just not have iwconfig installed?

tyll:
> @tyll Line 1265 tests for the `sort` command; which was used in that block. This patch moves its use outside of that block. It needs fixed somehow. :)

Well spotted . Line 1490 uses `sort` without a test for it. Also, the script uses `cat` multiple times without checking for it. `sort` and `cat` are both in coreutils: https://github.com/coreutils/coreutils/tree/master/src - would you be fine with removing the test from the if clause? Also `ls` is part of coreutils.

ZeroChaos-:
> would you be fine with removing the test from the if clause? Also `ls` is part of coreutils.

absolutely not.  the tests for coreutils may seem silly on your laptop, but they are required for things like openwrt

tyll:
> I finally had time to review this, and the whole thing makes no sense. It doesn't make sense to run through sort -u if we didn't get devices from two different sources which may provide duplicates. I agree this needs to be fixed, but not in the way which it's being fixed.
> 
> Is the simple test case to just not have iwconfig installed?

In my case, it broke with two wifi interfaces and no iwconfig. I guess it should trigger it.  Due to sort being part of coreutils I suppose running sort unnecessarily is not that harmful.

tyll:
> > would you be fine with removing the test from the if clause? Also `ls` is part of coreutils.
> 
> absolutely not. the tests for coreutils may seem silly on your laptop, but they are required for things like openwrt

@ZeroChaos- I see. I updated the patch to run sort only when iwconfig and sort are present. If you have a better solution, I am happy to learn about it and close this PR. Thanks.

ZeroChaos-:
Does this fix your issue? https://github.com/aircrack-ng/aircrack-ng/commit/1b856395b23b456f828be52fce82b42e2d35a32b

tyll:
> Does this fix your issue? [1b85639](https://github.com/aircrack-ng/aircrack-ng/commit/1b856395b23b456f828be52fce82b42e2d35a32b)

I guess it will fix my issue but break the deduplication of interfaces since `sort` sorts lines and not words:

```
printf "%s" "b a a " | sort -bu
b a a 
```

ZeroChaos-:
ha, excellent point, I'll fix that today.  I only tested without iwconfig installed, my bad

ZeroChaos-:
fixed, thanks for the report

PR #2318, opened 2022-06-30
airmon-ng: avoid direct call of systemctl command

zOrg1331:
Commit f94a3fe3 introduced systemd support additionally to SystemV. However,
this caused the additional dependency on systemctl executable which is not
desired in some cases. This commit puts it inside a variable which tricks
requirements calculation scripts (if such are used) to ignore systemd.

### DIFF
--- a/scripts/airmon-ng.linux
+++ b/scripts/airmon-ng.linux
@@ -1168,10 +1168,11 @@ scanProcesses() {
 
 	if [ "${1}" = "kill" ]; then
     if [ -x "$(command -v systemctl 2>&1)" ]; then
+        systemctlcmd="$(command -v systemctl 2>&1)"
 		  for service in network-manager NetworkManager avahi-daemon wicd; do
-        if systemctl status "${service}" > /dev/null 2>&1; then
+        if ${systemctlcmd} status "${service}" > /dev/null 2>&1; then
           killservice=1
-          until systemctl stop "${service}" 2> /dev/null > /dev/null; do
+          until ${systemctlcmd} stop "${service}" 2> /dev/null > /dev/null; do
             killservice=$((killservice + 1))
             if [ ${killservice} -gt 5 ]; then
               printf "Failed to stop %s, please stop it on your own.\n" "${service}"


ZeroChaos-:
This already has a check for the executable, what is the point of your change?

zOrg1331:
> This already has a check for the executable, what is the point of your change?

It indeed looks weird, I agree. I will try to explain.

What the change achieves is the absence of explicit command execution by this script. Instead of doing `systemctl` we do `${systemctlcmd}` where `$systemctlcmd` contains a path to the command. This does not change anything to the execution flow. However, the change emphasizes the fact that `systemctl` command is optional.

From Linux distro point of view, direct call of a command in a script means that the script depends on this command. This is quite important if the command is `systemctl` as the whole Systemd is required.

In the distro where I participate (ALTLinux) we have automated requirements checker which stumbles across such cases. We still try to separate Systemd and SystemV worlds. This change helps to achieve such separation by avoiding strict dependency on `systemctl`. Debian team does not have/use such automation and set requirements manually, which is handy, but could miss some corner-cases.

I hope I made it slightly more clear. :-D Or may be not.

ZeroChaos-:
> > This already has a check for the executable, what is the point of your change?
> 
> It indeed looks weird, I agree. I will try to explain.
> 
> What the change achieves is the absence of explicit command execution by this script. Instead of doing `systemctl` we do `${systemctlcmd}` where `$systemctlcmd` contains a path to the command. This does not change anything to the execution flow. However, the change emphasizes the fact that `systemctl` command is optional.
> 
It's already optional and ignored if it's not there https://github.com/aircrack-ng/aircrack-ng/pull/2318/files#diff-0949765fa844ae6201aeac870d50628269e0cd0885394a11f8c97c54ba9a59c4L1170
> From Linux distro point of view, direct call of a command in a script means that the script depends on this command. This is quite important if the command is `systemctl` as the whole Systemd is required.

Yeah, except there is a check for the executable, and it's not required.
> 
> In the distro where I participate (ALTLinux) we have automated requirements checker which stumbles across such cases. We still try to separate Systemd and SystemV worlds. This change helps to achieve such separation by avoiding strict dependency on `systemctl`. Debian team does not have/use such automation and set requirements manually, which is handy, but could miss some corner-cases.
> 
> I hope I made it slightly more clear. :-D Or may be not.

Just fyi, this is broken.  You have introduced a new variable which isn't set anywhere and breaks the code for literally everyone but you.  NACK for all of the above reasons.

PR #2336, opened 2022-10-14
added support for rtl8723cs

Shubhamvis98:
this chipset is used in pinephone(not pro). right now, if we use airmon-ng on pinephone to put it into monitor mode then monitor mode and packet injection won't work but if we use "iwconfig wlan0 mode monitor" then both monitor mode and packet injection work.

### DIFF
--- a/scripts/airmon-ng.linux
+++ b/scripts/airmon-ng.linux
@@ -5,6 +5,7 @@ ELITE="0"
 USERID=""
 checkvm_status=""
 MAC80211=0
+DEPRECATED=0
 IW_SOURCE="https://mirrors.edge.kernel.org/pub/software/network/iw/iw-5.9.tar.xz"
 IW_ERROR=""
 if [ ! -d '/sys/' ]; then
@@ -49,7 +50,7 @@ if [ -n "${3}" ];then
 		CH=3
 	fi
 else
-	CH=10
+	CH=3
 fi
 
 #TODO LIST
@@ -846,6 +847,7 @@ getDriver() {
 	fi
 	if [ "$DRIVER" = "rtl8723cs" ]; then
 		DRIVER="8723cs"
+		DEPRECATED=1
 	fi
 
 	#Here we will catch the broken lying drivers not caught above
@@ -1616,7 +1618,7 @@ for iface in $(printf "%s" "${iface_list}"); do
 		if [ "${1}" = "stop" ] && [ "${2}" = "${iface}" ]; then
 			stopwlIface "${iface}"
 		fi
-	elif [ "${DRIVER}" = "8723cs" ]; then
+	elif [ "${DEPRECATED}" = "1" ]; then
 		if [ "${1}" = "start" ] && [ "${2}" = "${iface}" ]; then
 			startDeprecatedIface "${iface}"
 		fi
--- a/scripts/airmon-ng.linux
+++ b/scripts/airmon-ng.linux
@@ -368,6 +368,12 @@ ifaceExists() {
 }
 
 startDeprecatedIface() {
+	if iwconfig "${1}" | grep -q 'Monitor'
+	then
+		printf "\t\t(Monitor mode already enabled for [%s]%s)\n" "${PHYDEV}" "${1}"
+		return
+	fi
+	setLink "${1}" down
 	iwconfig "${1}" mode monitor > /dev/null 2>&1
 	if [ -n "${2}" ]; then
 		if [ "${2}" -lt 1000 ]; then
@@ -649,6 +655,11 @@ setChannelMac80211() {
 }
 
 stopDeprecatedIface() {
+	if iwconfig "${1}" | grep -q 'Managed'
+	then
+		printf "\t\t(Monitor mode already disabled for [%s]%s)\n" "${PHYDEV}" "${1}"
+		return
+	fi
 	setLink "${1}" down
 	iwconfig "${1}" mode Managed > /dev/null 2>&1
 	setLink "${1}" up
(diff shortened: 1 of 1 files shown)


scripts/airmon-ng.linux
```diff
@@ -49,7 +50,7 @@ if [ -n "${3}" ];then
 		CH=3
 	fi
 else
-	CH=10
+	CH=3
```
Mister-X-, line 53: Why is the channel changed?

scripts/airmon-ng.linux
```diff
@@ -1616,7 +1618,7 @@ for iface in $(printf "%s" "${iface_list}"); do
 		if [ "${1}" = "stop" ] && [ "${2}" = "${iface}" ]; then
 			stopwlIface "${iface}"
 		fi
-	elif [ "${DRIVER}" = "8723cs" ]; then
+	elif [ "${DEPRECATED}" = "1" ]; then
```
Mister-X-, line 1621: Please include the driver check as well, in the event ${DEPRECATED} gets used with other drivers.

scripts/airmon-ng.linux
Shubhamvis98: Hi, if we use airmon-ng in pinephone without mentioning the channel number then it's changed to ch10 according to the shell script and if we switch to ch10 in rtl8723 then packet injection and mon mode are not working. I know this sounds strange but this driver has some issues. that's why I changed the channel number also.

scripts/airmon-ng.linux
Shubhamvis98: elif [ "${DEVICEID}" = '0x024c:0xb703' ]; then
  CHIPSET='Realtek RTL8723CS'

in line 988, this will check for the driver also. anything else I need to mention?

ZeroChaos-:
NACK, this is yucky.  I will take a stab at adding it right.

ZeroChaos-:
Thanks for the information, but this implementation is not going to be okay.

ZeroChaos-:
@Shubhamvis98 thanks for the patience.  I want to support you and your use case, but I've been busy.  I added the bits of this code which made sense, as well as a few other things to improve safety and documentation in the code.

At this time, I'm not willing to actually hook up the DeprecatedIface commands.  Nothing needs them, and nothing should.  I realize iwconfig is just begging to be used, but it's been deprecated for a decade longer than that wireless chipset you are using has existed.  I suspect, that instead of startDeprecatedIface and stopDeprecatedIface that changeMac80211IfaceTypeMonitor and changeMac80211IfaceTypeManaged should work. Can you please test using those instead of the deprecated ones?

ZeroChaos- (changes requested):
this PR is not going to be merged.  but I will take action based on the suggestions and based on the conversation.

Shubhamvis98:
> @Shubhamvis98 thanks for the patience. I want to support you and your use case, but I've been busy. I added the bits of this code which made sense, as well as a few other things to improve safety and documentation in the code.
> 
> At this time, I'm not willing to actually hook up the DeprecatedIface commands. Nothing needs them, and nothing should. I realize iwconfig is just begging to be used, but it's been deprecated for a decade longer than that wireless chipset you are using has existed. I suspect, that instead of startDeprecatedIface and stopDeprecatedIface that changeMac80211IfaceTypeMonitor and changeMac80211IfaceTypeManaged should work. Can you please test using those instead of the deprecated ones?

Thanks @ZeroChaos-, I'll try and let you know if it works or not. The driver of rtl8723cs is also behaving strange while putting it to mon mode. But as per your suggestion, I'll try. Thanks again.

Shubhamvis98:
> > @Shubhamvis98 thanks for the patience. I want to support you and your use case, but I've been busy. I added the bits of this code which made sense, as well as a few other things to improve safety and documentation in the code.
> > At this time, I'm not willing to actually hook up the DeprecatedIface commands. Nothing needs them, and nothing should. I realize iwconfig is just begging to be used, but it's been deprecated for a decade longer than that wireless chipset you are using has existed. I suspect, that instead of startDeprecatedIface and stopDeprecatedIface that changeMac80211IfaceTypeMonitor and changeMac80211IfaceTypeManaged should work. Can you please test using those instead of the deprecated ones?
> 
> Thanks @ZeroChaos-, I'll try and let you know if it works or not. The driver of rtl8723cs is also behaving strange while putting it to mon mode. But as per your suggestion, I'll try. Thanks again.

Unfortunately it's not working. The packet injection doesn't work if I use "iw" to change it to monitor mode.

ZeroChaos-:
> Unfortunately it's not working. The packet injection doesn't work if I use "iw" to change it to monitor mode.

just to confirm, it does work with iwconfig?

Also can you please provide the output of this command?
`iw phy phy0 info | grep 'interface combinations are not supported'`

Shubhamvis98:
> 

Yes, using iwconfig packet injection and monitor mode both works fine. 
iw

steev:
which rtl8723cs driver are you using?

Shubhamvis98:
Pinephone (non-pro) internal wifi.

steev:
That's the chipset, i'm asking what *driver* are you using; where is it from.

Shubhamvis98:
It's already included in mobian's kernel. From megi's kernel patch.

Shubhamvis98:
https://salsa.debian.org/Mobian-team/devices/kernels/sunxi64-linux/-/tree/mobian-6.1/debian/patches/wireless

The same kernel is also installed in kali-nethunter for pinephone official release

steev:
Okay, I use https://github.com/steev/rtl8723cs/commits/new-driver-by-megous for the kali pinebook (og), and do not see these issues at all, so I am assuming it's a problem in the driver, and I don't think the proper fix is random changes to aircrack, but instead to fix the driver.

Shubhamvis98:
Thanks @steev, I'll compile and check the mentioned driver in pinephone and let you know. Thank you for your time 'n' help.

ZeroChaos-:
Aside from the broken driver (working one mentioned by steev) this should all work correctly.  I really appreciate the suggested changes, and I've merged as many as I plan to.  As such, I'm closing this PR.  Thanks again for the work, I hope the recommended driver treats you well.

PR #2371, opened 2023-01-14
airdecloak-ng: fix typos and indentation in usage info

gemesa:
Before:
```bash
$ ./airdecloak-ng --help
...
           signal_dup_consec_sn: Use signal (if available), duplicate and
                                 consecutive sequence number (filtering is
                                  much more precise than using all these
                                  filters one by one).
     --null-packets        : Assume that null packets can be cloaked.
     --disable-base_filter : Do not apply base filter.
     --drop-frag           : Drop fragmented packets

     --help                : Displays this usage screen
```
     
After:
```bash
$ ./airdecloak-ng --help
...
           signal_dup_consec_sn: Use signal (if available), duplicate and
                                 consecutive sequence number (filtering is
                                 much more precise than using all these
                                 filters one by one).
     --null-packets        : Assume that null packets can be cloaked.
     --disable-base-filter : Do not apply base filter.
     --drop-frag           : Drop fragmented packets.

     --help                : Displays this usage screen.
```

### DIFF
--- a/src/airdecloak-ng/airdecloak-ng.c
+++ b/src/airdecloak-ng/airdecloak-ng.c
@@ -1532,17 +1532,17 @@ static void usage(void)
 		"and\n"
 		"                                 consecutive sequence number "
 		"(filtering is\n"
-		"                                  much more precise than using all "
+		"                                 much more precise than using all "
 		"these\n"
-		"                                  filters one by one).\n"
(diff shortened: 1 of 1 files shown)


gemesa:
[wiki](https://www.aircrack-ng.org/doku.php?id=airdecloak-ng) has been updated accordingly.

PR #2379, opened 2023-01-15
airdecloak-ng, libac/support/common: fix comment and print related typos

gemesa:


src/airdecloak-ng/airdecloak-ng.c
```diff
@@ -560,7 +560,7 @@ static BOOLEAN read_packets(void)
 		}
 
 #ifdef DEBUG
-		printf("From DS: %d - ToDS: %d\n",
+		printf("From DS: %d - To DS: %d\n",
```
Mister-X-: Actually, ToDS is correct, and that's "From DS" that should be changed to "FromDS"

Mister-X-: Actually, ToDS is correct, and that's "From DS" that should be changed to "FromDS"

gemesa: Right, fixed it.

PR #2385, opened 2023-01-20
add PCRE2 support

### LINKED ISSUE #2323: Upgrading from pcre to pcre2
# Description

The pcre library is deprecated and no longer supported by the upstream. Please consider switching the project to [pcre2](https://github.com/PCRE2Project/pcre2).

gemesa:
Fixes #2323 

I think we should switch from PCRE to PCRE2 in the long term as PCRE is deprecated and abandoned since 2015. For now I added support for PCRE2 and kept the PCRE code also.

Looking for files using PCRE:
```bash
┌───(gemesa♾fedora)-[~/git-repos/aircrack-ng-sync]
└─$ grep -rnw . -e 'pcre.h'
./src/airodump-ng/airodump-ng.c:72:#include <pcre.h>
./src/airodump-ng/dump_write.c:49:#include <pcre.h>
./src/besside-ng/besside-ng.c:61:#include <pcre.h>
```

_airodump-ng.c_ and _besside-ng.c_ are really utilizing PCRE, _dump_write.c_ just includes it but does not use it.

The C code I used to validate the integrated PCRE2 code:
```C
#include <stdio.h>
#include <string.h>

#define ESSID_LENGTH 32

#define PCRE2_CODE_UNIT_WIDTH 8
#include <pcre2.h>

#include <getopt.h>

int main(int argc, char * argv[])
{
    char * essid = NULL;
    char * regex = NULL;
    pcre2_code * f_essid_regex;
    pcre2_match_data * f_essid_match_data;
    int pcreerror;
    PCRE2_SIZE pcreerroffset;
    int option;
    int match;

    while ((option = getopt(argc, argv, "N:R:")) != -1)
    {
        switch (option)
        {
            case 'N':
                essid = malloc(strlen(optarg));
                if (essid == NULL)
                {
                    puts("malloc error");
                    return EXIT_FAILURE;
                }
                strcpy(essid, optarg);
                break;
            case 'R':
                regex = malloc(strlen(optarg));
                if (regex == NULL)
                {
                    puts("malloc error");
                    return EXIT_FAILURE;
                }
                strcpy(regex, optarg);
                break;
            default:
                puts("unsupported argument");
                return EXIT_FAILURE;
        }
    }

    if ((essid == NULL) || ((regex == NULL)))
    {
        puts("Please provide -N and -R options");
        return EXIT_FAILURE;
    }

    f_essid_regex
        = pcre2_compile((PCRE2_SPTR)regex, PCRE2_ZERO_TERMINATED, 0, &pcreerror, &pcreerroffset, NULL);
    
    if (f_essid_regex == NULL)
    {
        PCRE2_UCHAR pcreerrbuffer[256];
        pcre2_get_error_message(pcreerror, pcreerrbuffer, sizeof(pcreerrbuffer));

        printf("Error: regular expression compilation failed at "
               "offset %lu: %s; aborting\n",
               pcreerroffset,
               pcreerrbuffer);

        exit(EXIT_FAILURE);
    }
    
    f_essid_match_data
        = pcre2_match_data_create_from_pattern(f_essid_regex, NULL);

    match =  pcre2_match(f_essid_regex,
                        (PCRE2_SPTR) essid,
                        (int) strnlen((char *) essid, ESSID_LENGTH),
                        0,
                        0,
                        f_essid_match_data,
                        0);

    printf("match result: %d\n", match);

    pcre2_match_data_free(f_essid_match_data);
    pcre2_code_free(f_essid_regex);
    
    free(essid);
    free(regex);

    return EXIT_SUCCESS;
}
```

```bash
$ gcc pcre2-test.c -o pcre2-test -lpcre2-8
$ ./pcre2-test -N "" -R ".+"                              
match result: -1
$ ./pcre2-test -N "1" -R ".+"
match result: 1
$ ./pcre2-test -N "12a34" -R "\\d{2}\\w\\d{2}"
match result: 1
$ ./pcre2-test -N "12a34" -R "\\d{2}\\d\\d{2}"
match result: -1
$ ./pcre2-test -N "12a34" -R "*"              
Error: regular expression compilation failed at offset 0: quantifier does not follow a repeatable item; aborting
$ ./pcre2-test -N "12a34" -R "("
Error: regular expression compilation failed at offset 1: missing closing parenthesis; aborting
```

Relevant sources:
- https://github.com/aircrack-ng/aircrack-ng/commit/bd84e9888eefa4d3d22a7b7720b6b71423194472
- https://github.com/i3/i3/issues/4682
- https://github.com/psychon/i3/commit/2defec9c9c9a65025cabbcf673ca608f077835d3
- https://github.com/PCRE2Project/pcre2
- https://github.com/PCRE2Project/pcre2/blob/master/src/pcre2test.c
- https://github.com/luvit/pcre2/blob/master/src/pcre2demo.c


build/m4/aircrack_ng_pcre2.m4
```diff
@@ -0,0 +1,61 @@
+dnl Aircrack-ng
+dnl
+dnl Copyright (C) 2017 Joseph Benden <joe@benden.us>
```
Mister-X-: Don't forget to adjust the copyrights

gemesa: Right, should I update the date only or put my name there also?

Mister-X-:
I see that some of the API is very similar to the old PCRE. What are your thoughts about having a compat-type header file with a few `#define`, kinda like what we had for openssl and gcrypt, for some of the functions? Do you see what I mean?

I believe that should reduce the size of the code in the `#ifdef` a little bit and make it more readable.

gemesa:
> I see that some of the API is very similar to the old PCRE. What are your thoughts about having a compat-type header file with a few `#define`, kinda like what we had for openssl and gcrypt, for some of the functions?

You mean for example the different definitions of `Cipher_AES_Encrypt_Init()`, `Cipher_AES_Encrypt()` and `Cipher_AES_Encrypt_Deinit()` in _aes-128-cbc-openssl.c_ and _aes-128-cbc-gcrypt.c_?

So in case of _airodump-ng.c_ and _besside-ng.c_ I could move for example `is_filtered_essid()` to separate _.c_ files like _airodump-ng-pcre.c_, _airodump-ng-pcre2.c_ and _besside-ng-pcre.c_, _besside-ng-pcre2.c_? It would make sense imo if that is what you mean.

build/m4/aircrack_ng_pcre2.m4
gemesa: fixed

Mister-X-:
Here is a practical example of what I'm thinking about:

```c
#define COMPAT_PCRE_MATCH(regex, essid, match_data) \
	#ifdef HAVE_PCRE2 \
		pcre2_match((regex), \
				(PCRE2_SPTR) (essid), \
				(int) strnlen((char *) (essid), MAX_IE_ELEMENT_SIZE), \
				0, \
				0, \
				(match_data), \
				0)
	#elif defined HAVE_PCRE \
		pcre_exec((regex), \
				NULL, \
				(char *) (essid), \
				strnlen((char *) (essid), MAX_IE_ELEMENT_SIZE), \
				0, \
				0, \
				NULL, \
				0) \
	#endif
```

This [section](https://github.com/aircrack-ng/aircrack-ng/pull/2385/files#diff-3cee5e6de0ecb02b103ef57b4f747cc04b046ab8f4ed2c0d6f2fb2e396812311R1125-R1171) is pretty much duplicate code, where there is just one function that is different. So that `pcre_exec` could be changed to `return COMPAT_PCRE_MATCH(_conf.cf_essid_regex, essid, _conf.cf_essid_match_data) < 0;`

gemesa:
Well generally I dont like function-like macros but I see the benefits in this case and also duplicated code is bad for sure so I will rework the code accordingly.

gemesa:
I rewrote it, please check it and let me know what you think.

Mister-X-:
Armel buildbot indicates [2 identical warnings](https://buildbot.aircrack-ng.org/#/builders/8/builds/1258/steps/4/logs/warnings__2_) for the build:

```
/buildbot/local/aircrack-ng-armel/build/include/aircrack-ng/pcre/compat-pcre.h:79:16: warning: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 2 has type ‘size_t’ {aka ‘unsigned int’} [-Wformat=]
```

It seems like the correct flag would be `%zu` instead of `%lu`, as [`PCRE2_SIZE`](https://pcre.org/current/doc/html/pcre2limits.html) is defined as a `size_t`

gemesa:
> It seems like the correct flag would be `%zu` instead of `%lu`, as [`PCRE2_SIZE`](https://pcre.org/current/doc/html/pcre2limits.html) is defined as a `size_t`

You are right, fixed it.

gemesa:
I built and tested `airodump-ng` (censored some data manually):

```
$ ./airodump-ng mon0 --essid-regex "\w+-\d+"

CH  8 ][ Elapsed: 2 mins ][ 2023-01-29 12:59

 BSSID              PWR  Beacons    #Data, #/s  CH   MB   ENC CIPHER  AUTH ESSID

 xx:xx:xx:xx:xx:xx  -91        6        0    0   1  195   WPA2 CCMP   PSK  Vodafone-xxxx
 xx:xx:xx:xx:xx:xx  -85        1        2    0   1  540   WPA2 CCMP   PSK  Telekom-xxxxxx
 xx:xx:xx:xx:xx:xx  -82        0        2    0   1  195   WPA2 CCMP   PSK  Vodafone-xxxx
 xx:xx:xx:xx:xx:xx  -87        7        0    0  11  130   WPA2 CCMP   PSK  Vodafone-xxxx
 xx:xx:xx:xx:xx:xx  -73       85        0    0  11  130   WPA2 CCMP   PSK  Vodafone-xxxx
 xx:xx:xx:xx:xx:xx  -87       10        0    0  13  195   WPA2 CCMP   PSK  Telekom-xxxxxx
 xx:xx:xx:xx:xx:xx  -82       13        4    0   6  130   WPA2 CCMP   PSK  Vodafone-xxxx
 xx:xx:xx:xx:xx:xx  -87       36        0    0  11  195   WPA2 CCMP   PSK  Vodafone-xxxx
 xx:xx:xx:xx:xx:xx  -88        7        0    0  11  540   WPA3 CCMP   SAE  Telekom-xxxxxx
 xx:xx:xx:xx:xx:xx  -90        9        0    0  11  195   WPA2 CCMP   PSK  Vodafone-xxxx
 xx:xx:xx:xx:xx:xx  -89       15        1    0   1  540   WPA3 CCMP   SAE  Telekom-xxxxxx
 xx:xx:xx:xx:xx:xx  -83       23        0    0   1  130   WPA2 CCMP   PSK  Vodafone-xxxx
 xx:xx:xx:xx:xx:xx  -62       81        0    0   1  195   WPA2 CCMP   PSK  Vodafone-xxxx
```

PR #2403, opened 2023-01-24
workflows: fix gcc and clang in macos.yml

### LINKED ISSUE #2376: Fix MacOS GitHub action
# Defect Report

## Issue type

- [x] Defect - Compilation or `make check` issue - Attach `config.log` and provide relevant system information such as `lscpu` - Make sure dependencies are installed

## System information

- OS Name: MacOS
- CPU: i386

## Aircrack-ng version

- Version: 1.7
- Commit Revision hash:  e781834

## Airmon-ng debug information

N/A

### Details

The action fails with gcc-9 and gcc-10.

At least with gcc-10, we can see the error `ccache: error: Could not find compiler "gcc-10" in PATH` in the logs. See https://github.com/aircrack-ng/aircrack-ng/actions/runs/3850459568/jobs/6560625246 for example

### How to reproduce the issue

Commit to the repository

## Related issues

N/A

gemesa:
Fixes #2376 

Added `gcc-9` (`gcc@9`) and `gcc-10` (`gcc@10`) to the install dependencies to fix `gcc-9` and `gcc-10`.

After enabling `clang-11`, it was necessary to add missing void keywords to some function definitions and to comment out some unused but set variables to fix `clang-11`. These were necessary because `clang-11` on MacOs is very strict for some reason and raises -Wunused-but-set-variable and can not match the function declaration and the function definition if the void keyword is missing from the argument, for example:

```
void init_atoi(void);
void init_atoi()
{
 ...
}
```

Since I have no Apple HW so I can not test building myself, I have been using the MacOs runner to do that for me. This generated a lot of junk action which can be removed. Sorry about that.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -32,10 +32,10 @@ jobs:
         compiler:
           - cc:  gcc-11
             cxx: g++-11
-          #- cc:  gcc-10
-          #  cxx: g++-10
-          #- cc:  gcc-9
-          #  cxx: g++-9
+          - cc:  gcc-10
+            cxx: g++-10
+          - cc:  gcc-9
+            cxx: g++-9
           #- cc:  clang-11     broken handling with version.h
           #  cxx: clang++-11
           - cc:  clang-12
@@ -65,7 +65,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          brew install autoconf automake ccache cmocka expect hwloc libpcap libtool openssl pcre2 pkg-config sqlite3 shtool md5sha1sum
+          brew install gcc@9 gcc@10 autoconf automake ccache cmocka expect hwloc libpcap libtool openssl pcre2 pkg-config sqlite3 shtool md5sha1sum
 
       - name: Ccache stats before builds
         run: |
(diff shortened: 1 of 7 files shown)


Mister-X-:
This deletes a bunch of workflows and do changes unrelated to this PR (Check "Files" tab in this PR).

Mister-X-:
clang-11 had an issue with version.h, is that resolved?

By the way, GitHub says there is a conflict with the now current `macos.yml` CI

gemesa:
> clang-11 had an issue with version.h, is that resolved?

I could not locate any information about this issue because the initial version (https://github.com/aircrack-ng/aircrack-ng/commit/2beb3e2341bcf19d0602e59c1e618e801d69bc52) of this workflow already contained `clang-11` commented out. But since it is building without errors I think we can conclude it is solved.

> By the way, GitHub says there is a conflict with the now current `macos.yml` CI

Merge conflict resolved.

gemesa:
I am going to squash the merge commit.

gemesa:
Merge conflict resolved without a merge commit so the 2 commits can be merged cleanly if you agree with their content.

Issue #1940, opened 2018-08-11
Filter multiple MAC addresses


aircrack-ng:
Ability to filter multiple MAC address with no common OUI mask

drws:
One way would be to support passing multiple `--bssid` arguments and the same goes for other filtering options (`--encrypt`, `--netmask`, `--essid`, `--essid-regex`).- Alternatively, `airodump-ng` could be able to take a file with multiple filters as an input (which would require adding `--bssid-list`, `--encrypt-list`, ...).

drws:
@Mister-X- What about passing multiple `--encrypt`, `--netmask`, `--essid`, `--essid-regex`? Should that be moved to a new issue?

Mister-X-:
For regex, I don't see it being multiple. You just have to adjust your regex to match more SSID. Same goes for SSID, we can use a regex for that.

However, for encryption, yes, that would make sense to have a separate issue opened for it. Make sure to mention this issue in it.

Feel free to open one for netmask, but that's gonna be a discussion as it depends on another parameter, so we need to think about a good solution.

drws:
I already had the `--essid-regex` removed from that list, but then I thought it could still be useful for prettifying the input sometimes. Since this would introduce redundant functionality I agree it is not really needed.

On the other hand I'd like to argue that `--essid-regex` is not truly a multiple `--essid` replacement, because of the following:
1. User has to be familiar with regex (or need to remember it) in contrast to passing multiple `--essid`.
2. `--essid-regex` introduces additinal characters in the filter string, masking and confusing the actual SSID data.
3. `--essid-regex` requires escaping more characters than `--essid`.
4. Since multiple `--bssid`s are supported now, user can rightfully expect a consistent behaviour with multi-`--essid`s.

For `--encrypt` a new issue was created.

And lastly, regarding `--netmask`. I'm leaving it up to you since I don't know about the corresponding issue(s) you mentioned.

Mister-X-:
OK, you have a good point. Go ahead with a new ticket for `--essid`.

For the --netmask, start a [discussion](https://github.com/aircrack-ng/aircrack-ng/discussions), as we need to figure out the right way to implement this as this depends upon another parameter

drws:
Regarding `--netmask` I'm still unsure which parameter dependence would require a discussion? The documented relation to `--bssid` or something else?

Mister-X-:
For now, just `--bssid` AFAIK.

drws:
[...]

PR #2428, opened 2023-02-10
airbase-ng: remove --netmask related dead code

gemesa:
Removed `--netmask` related dead code from `airbase-ng` as `airbase-ng` has no such option, probably a copy-paste error. Noticed while working on #1940 and #2432.


gemesa:
Are you interested in this change @Mister-X- ?

Mister-X-:
Yes. I just need to take more time to look in the code.

PR #2474, opened 2023-02-27
workflows: add missing apparmor/** and manpage/** to ignore paths

gemesa:
While working on #2468 I noticed modifying manpages also triggers some build workflows (Kali, Alpine etc.). This is because `manpage/**` is not added to the ignore paths in some CI `.yml` files. Now this is fixed, and `apparmor/**` is added as well.


Mister-X-:
For Docker that pushes to DockerHub, let's just ignore the apparmor profiles, as the manpages get added to the containers.

gemesa:
> For Docker that pushes to DockerHub, let's just ignore the apparmor profiles, as the manpages get added to the containers.

Updated accordingly.

PR #2488, opened 2023-03-03
airodump-ng: fix displaying -M and -W columns

gemesa:
When `--manufacturer` option is provided to `airodump-ng` the column titles are not displayed properly:

```
$ sudo airodump-ng wlp0s20f0u1 --manufacturer --wps

 CH  2 ][ Elapsed: 0 s ][ 2023-03-03 20:32 ][ Are you sure you want to quit? Press Q again to quit.

 BSSID              PWR  Beacons    #Data, #/s  CH   MB   ENC CIPHER  AUTH WPS    ESSI          MANUFACTURER

 XX:XX:XX:XX:XX:XX  -78        2        0    0   9  130   WPA2 CCMP   PSK  2.0    XXXXXXXXXXXXX Compal Broadband Networks, Inc.
```

As you can see `ESSI` is displayed instead of `ESSID`.

```
$ sudo airodump-ng wlp0s20f0u1 --manufacturer

 CH  7 ][ Elapsed: 0 s ][ 2023-03-03 20:34 ][ Are you sure you want to quit? Press Q again to quit.

 BSSID              PWR  Beacons    #Data, #/s  CH   MB   ENC CIPHER  AUTH ESSI        MANUFACTURER

 XX:XX:XX:XX:XX:XX  -83        2        0    0   1  130   WPA2 CCMP   PSK  XXXXXXX      Compal Broadband Networks, Inc.
```

As you can see `ESSI` is displayed instead of `ESSID` and `MANUFACTURER` is shifted by one space.

These issues have multiple reasons: `columns_ap` was hardcoded to `83` but the length of this line is `84`:
```
 BSSID              PWR  Beacons    #Data, #/s  CH   MB   ENC CIPHER  AUTH ESSID
```
Also, there were some magic numbers going on (`-1`, `-5`, `-7`, `32`) and some of them were wrong. Not to mention they made the code difficult to read. Now these issues have been fixed.

Test measurements:

```
$ sudo ./airodump-ng wlp0s20f0u1 --wps --manufacturer

 CH 13 ][ Elapsed: 0 s ][ 2023-03-03 20:42 ][ Are you sure you want to quit? Press Q again to quit.

 BSSID              PWR  Beacons    #Data, #/s  CH   MB   ENC CIPHER  AUTH WPS    ESSID         MANUFACTURER

 XX:XX:XX:XX:XX:XX  -57        3        0    0   6  195   WPA2 CCMP   PSK Locked  XXXXXXXXXXXXX ARRIS Group, Inc.
```

[...]

PR #2489, opened 2023-03-03
airodump-ng: replace number 32 with ' ' (space)

gemesa:
This is a follow-up of #2488 

Every remaining occurrence of magic number `32` has been replaced with `' '` (space).


gemesa:
This could be all greens also. Im sad because the BSD CIs are a bit unstable but it is what it is. Maybe they will improve in the future.

gemesa:
> This could be all greens also. Im sad because the BSD CIs are a bit unstable but it is what it is. Maybe they will improve in the future.

There is a good chance this can be fixed with #2490.

PR #2494, opened 2023-03-04
autogen.sh: fix shellcheck findings

gemesa:
Fixed `shellcheck` findings in `autogen.sh`.

`shellcheck` version:
```
$ shellcheck -V
ShellCheck - shell script analysis tool
version: 0.8.0
license: GNU General Public License, version 3
website: https://www.shellcheck.net
```

Verifying `olddir` variable is unused:
```
$ grep -rni . -e 'olddir'
./autogen.sh:47:olddir="$(pwd)"
```

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -44,7 +44,6 @@ set -euf
 test -n "${srcdir:-}" || srcdir="$(dirname "$0")"
 test -n "${srcdir:-}" || srcdir=.
 
-olddir="$(pwd)"
 cd "$srcdir"
 
 missing_tool()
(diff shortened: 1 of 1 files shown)


Mister-X-:
What does `:-` do at the end of the string?

gemesa:
`:` is used to specify a default value if the variable is not set. So if `ACLOCAL_FLAGS` has a value assigned to it, that value will be returned, otherwise an empty string will be returned. Difference between `-` and `=` is that `-` does not set the variable to the default value but `=` does.

gemesa:
Usually it is something like `${ACLOCAL_FLAGS:-default_value}`. I dont know why is it like this here.

PR #2167, opened 2020-07-26
airodump-ng: add manufacturer column to the clients list

sighook:
Aircrack's airodump utility has an option --manufacturer:

    --manufacturer : Display manufacturer from IEEE OUI list

Currently, it shows only AP (`BSSID`) manufacturer, but will be come in handy to see also connected stations manufacturer.

For example, sometimes due pentesting purposes is necessary to view stations manufacturer and to not disconnecting all clients from the AP, but only one/two/etc (the most vulnerable from the point of view of Social Engineering).

For example.

Before patching:
```
 BSSID             STATION            PWR    Rate   Lost  Frames Notes Probes
 XX:XX:XX:XX:XX:XX XX:XX:XX:XX:XX:XX -54    0 - 0e  0     1            XXX,XXXXX,XXXXXX
 XX:XX:XX:XX:XX:XX XX:XX:XX:XX:XX:XX -38    0 - 0e  155   43
```
After:
```
 BSSID             STATION            PWR    Rate   Lost  Frames Notes Manufacturer                    Probes
 XX:XX:XX:XX:XX:XX XX:XX:XX:XX:XX:XX -54    0 - 0e  0     1            Hon Hai Precision Ind. Co.,Ltd. XXX,XXXXX,XXXXXX

 XX:XX:XX:XX:XX:XX XX:XX:XX:XX:XX:XX -38    0 - 0e  155   43           Liteon Technology Corporation
 ```

This PR adds manufacturer column to the clients list. It shows STATION's manufacturer.

Kind Regards,
A.

PR #2495, opened 2023-03-04
airodump-ng: refactor displaying of STATIONs

gemesa:
Did some refactoring in `airodump-ng`, this way the code can be extended easier if we want to add additional - variable length - columns. Main difference is that I replaced `printf()` with `snprintf()` so It is also consistent with the rest of the code.
This is a preparation for implementing #2167. I think that feature is a good idea and also got some attention by other users so I will reimplement it. That implementation has some bugs though, like the empty lines and does not handle the case properly if the manufacturer string is longer than 32 characters.

Fixed some indentation issues also in the displaying of stations.

Before:

```
$ sudo airodump-ng wlp0s20f0u1
 BSSID              STATION            PWR   Rate    Lost    Frames  Notes  Probes

 XX:XX:XX:XX:XX:XX  XX:XX:XX:XX:XX:XX  -55    0 - 6      0        1
```

As you can see `Rate` and `Lost` column titles are shifted by one space.

After:

```
$ sudo ./airodump-ng wlp0s20f0u1
 BSSID              STATION            PWR    Rate    Lost   Frames  Notes  Probes

 (not associated)   XX:XX:XX:XX:XX:XX  -87    0 - 1     18        6         XXXXXXXXXX
```

### DIFF
--- a/src/airodump-ng/airodump-ng.c
+++ b/src/airodump-ng/airodump-ng.c
@@ -4209,7 +4209,7 @@ static void dump_print(int ws_row, int ws_col, int if_num)
 	{
 		strlcpy(strbuf,
 				" BSSID              STATION "
-				"           PWR   Rate    Lost    Frames  Notes  Probes",
+				"           PWR    Rate    Lost   Frames  Notes  Probes",
 				sizeof(strbuf));
 		strbuf[ws_col - 1] = '\0';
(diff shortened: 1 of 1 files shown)


gemesa:
Closing for now as I have to make sure the UI keeps working properly. Will review it later.

Issue #2506, opened 2023-03-09
Airdecap-ng test #5 fails in some conditions


Mister-X-:
# Defect Report

## Issue type

- [x] Defect - Crash

## System information

- OS Name: Ubuntu/NetBSD
- OS Version: 22.04/9.3
- Other: gcc-10 with openssl/gcc 7.5 (maybe) with openssl
- CPU: x86-64
- Wireless card and chipset (if relevant): N/A

## Aircrack-ng version

- Version: 1.7
- Commit Revision hash:  eb59ba9

### Details

Airdecap-ng test #5 fails in these 2 cases:
- [Ubuntu](https://github.com/aircrack-ng/aircrack-ng/actions/runs/4377829787/jobs/7661806882)
- [NetBSD](https://github.com/aircrack-ng/aircrack-ng/actions/runs/4377829771/jobs/7661804928)

### How to reproduce the issue

Likely committing something to the repo.

PR #2522, opened 2023-03-15
workflows: fix build with clang* in linux CI

gemesa:
Fixed broken builds: 

[Ubuntu 22.04 clang-12 using Gcrypt](https://github.com/aircrack-ng/aircrack-ng/actions/runs/4420816213/jobs/7750961358#logs):

```
  configure:5175: checking for C compiler version
  configure:5184: ccache clang-11 --version >&5
  ccache: error: Could not find compiler "clang-11" in PATH
  configure:5195: $? = 1
```

[Ubuntu 22.04 clang-12 using OpenSSL](https://github.com/aircrack-ng/aircrack-ng/actions/runs/4420816213/jobs/7750961443#logs):
```
  configure:5175: checking for C compiler version
  configure:5184: ccache clang-11 --version >&5
  ccache: error: Could not find compiler "clang-11" in PATH
  configure:5195: $? = 1
```

Fixed `clang-13` and `clang-14` builds where the options meant for `gcc` are used currently (instead of `-Werror -Wno-zero-length-array -Wno-deprecated-declarations`):

[Ubuntu 22.04 clang-14 using OpenSSL](https://github.com/aircrack-ng/aircrack-ng/actions/runs/4420816213/jobs/7750961825#logs):
```
      C Compiler:                  ccache clang-14
      C++ Compiler:                ccache clang++-14 -std=gnu++17
      Python:                      python
  
      CFLAGS:                      -Werror -Wno-unused-result -Wno-deprecated-declarations
      CXXFLAGS:                    -Werror -Wno-unused-result -Wno-deprecated-declarations
```

[Ubuntu 22.04 clang-13 using OpenSSL](https://github.com/aircrack-ng/aircrack-ng/actions/runs/4420816213/jobs/7750961648#logs):
```
      C Compiler:                  ccache clang-13
      C++ Compiler:                ccache clang++-13 -std=gnu++17
      Python:                      python
  
      CFLAGS:                      -Werror -Wno-unused-result -Wno-deprecated-declarations
      CXXFLAGS:                    -Werror -Wno-unused-result -Wno-deprecated-declarations
```

PR #2523, opened 2023-03-15
airdecap-ng: make test-0005.sh more verbose

gemesa:
Follow-up of #2510

Made `test-airdecap-ng-0005.sh` more verbose, so we can debug #2506 further the next time we encounter the issue. 

In [jobs/7757968179](https://github.com/aircrack-ng/aircrack-ng/actions/runs/4424304425/jobs/7757968179) `test/test-airdecap-ng-0005.sh` failed again:
```
  Unexpected airdecap-ng output
  Decrypted file: /tmp/tmp.wcUz6ZXJUi
  FAIL test/test-airdecap-ng-0005.sh (exit status: 1)
```

It looks like `airdecap-ng` returned 0 but still the calculated hash does not match.


gemesa:
Maybe there is some corruption going on with the tmp files.

gemesa:
Build fails but is fixed in #2522

PR #2526, opened 2023-03-15
airdecap-ng: make test-0005.sh more verbose

gemesa:
Follow-up of #2523 where I print the name of the tmp file instead of its content:
```
echo "Actual MD5 hash: ${TMP_MD5}"
```

Should be:
```
echo -n "Actual MD5 hash: "; cat ${TMP_MD5}
```

We encountered though [this weird test failure](https://github.com/aircrack-ng/aircrack-ng/actions/runs/4429572802/jobs/7770205306) once again. I checked manually the expected output of `airdecap-ng`:

```
$ ./test/test-airdecap-ng-0005.sh
...
Total number of stations seen            1
Total number of packets read           139
Total number of WEP data packets         0
Total number of WPA data packets        46
Number of plaintext data packets         0
Number of decrypted WEP  packets         0
Number of corrupted WEP  packets         0
Number of decrypted WPA  packets        46
Number of bad TKIP (WPA) packets         0
Number of bad CCMP (WPA) packets         0
...
```

and this is what we get in the failed CI run:

```
Read 24 packets...
Total number of stations seen            1
Total number of packets read           139
Total number of WEP data packets         0
Total number of WPA data packets        46
Number of plaintext data packets         0
Number of decrypted WEP  packets         0
Number of corrupted WEP  packets         0
Number of decrypted WPA  packets        46
Number of bad TKIP (WPA) packets         0
Number of bad CCMP (WPA) packets         0
```

Notice that there is that extra line `Read 24 packets...`. Maybe this is what causes the MD5 mismatch. We can make sure next time we encounter this because then the calculated MD5 will be printed also in the CI.

### DIFF
--- a/test/test-airdecap-ng-0005.sh
+++ b/test/test-airdecap-ng-0005.sh
@@ -42,7 +42,7 @@ if [ "$(cat ${TMP_MD5})" != '45a93bc091a3929a7d63f86ddbb81401' ]; then
 	echo "Unexpected airdecap-ng output:"
 	echo "$airdecap_output"
 	echo "Expected MD5 hash: 45a93bc091a3929a7d63f86ddbb81401"
-	echo "Actual MD5 hash: ${TMP_MD5}"
+	echo -n "Actual MD5 hash: "; cat ${TMP_MD5}
 	echo "Decrypted file: ${TMP_DEC}"
 	exit 1
 fi


gemesa:
When that extra line is not printed the MD5 is OK:
```
...
Total number of stations seen            1
Total number of packets read           139
Total number of WEP data packets         0
Total number of WPA data packets        46
Number of plaintext data packets         0
Number of decrypted WEP  packets         0
Number of corrupted WEP  packets         0
Number of decrypted WPA  packets        46
Number of bad TKIP (WPA) packets         0
Number of bad CCMP (WPA) packets         0
Expected MD5 hash: 45a93bc091a3929a7d63f86ddbb81401
Actual MD5 hash: 45a93bc091a3929a7d63f86ddbb81401
Decrypted file: /tmp/tmp.10BrelMdY5
```

So we just need to figure out why it is printed randomly. This is the related part in `airdecap-ng.c`, I dont fully understand yet what is going on:

```
	/* loop reading and deciphering the packets */

	memset(&stats, 0, sizeof(stats));
	tt = time(NULL);

	while (1)
	{
		if (time(NULL) - tt > 0)
		{
			/* update the status line every second */

			erase_line(0);
			printf("Read %lu packets...\r", stats.nb_read);
			fflush(stdout);
			tt = time(NULL);
		}
...
```

Mister-X-:
I don't think this is a big deal. What is important are the numbers `airdecap-ng` outputs. But feel free to open a bug report if you want to keep track of it for later.

What I think we can do is just `grep -E '^(Total|Number) '` the output then do a MD5 of that.

gemesa:
> What I think we can do is just `grep -E '^(Total|Number) '` the output then do a MD5 of that.

I will do this in a follow-up PR and will remove the temporarily added debug code.

jbenden:
Program exited before that printf finishes, and the following statement that SHOULD of flushed the IO, didn't.... GDB might show it easier, if needed...

PR #2551, opened 2023-07-10
airmon-ng.linux: in a for loop, replace `ls` with a glob

wileyhy:
The `[ -e` syntax covers the case of when the glob produces no matches and `sh` uses the glob pattern itself as a value for "$i".

#### original
```sh
set -x; 
for i in $(ls /sys/class/ieee80211/); do 
  if [ ! -d "/sys/class/ieee80211/${i}/device/net" ]; then 
    MISSING_INTERFACE="${i}"; 
    echo "$MISSING_INTERFACE" :: "$i"; 
  fi; 
done; 
set - 
```

> \++ ls --color=auto /sys/class/ieee80211/
> \+ for i in $(ls /sys/class/ieee80211/)
> \+ '[' '!' -d /sys/class/ieee80211/phy0/device/net ']'
> \+ set -
> 

#### change added
```sh
set -x; 
for i in /sys/class/ieee80211/*; do 
  if [ -e "${i}" ] && [ ! -d "${i}/device/net" ]; then 
    MISSING_INTERFACE="${i}"; 
    echo "$MISSING_INTERFACE" :: "$i"; 
  fi; 
done; 
set -
```

> \+ for i in /sys/class/ieee80211/*
> \+ '[' -e /sys/class/ieee80211/phy0 ']'
> \+ '[' '!' -d /sys/class/ieee80211/phy0/device/net ']'
> \+ set -
> 

Edit 1: formatting
Edit 2: note re `[ -e` syntax

### DIFF
--- a/scripts/airmon-ng.linux
+++ b/scripts/airmon-ng.linux
@@ -203,8 +203,6 @@ usage() {
 handleLostPhys() {
   MISSING_INTERFACE=""
   if [ -d /sys/class/ieee80211 ]; then
-    #this should be fixed, but it's not going to be right now
-    # shellcheck disable=2045
     for i in /sys/class/ieee80211/*; do
       if [ -e "${i}" ] && [ ! -d "${i}/device/net" ]; then
         MISSING_INTERFACE="${i}"
--- a/scripts/airmon-ng.linux
+++ b/scripts/airmon-ng.linux
@@ -205,7 +205,7 @@ handleLostPhys() {
   if [ -d /sys/class/ieee80211 ]; then
     for i in /sys/class/ieee80211/*; do
       if [ -e "${i}" ] && [ ! -d "${i}/device/net" ]; then
-        MISSING_INTERFACE="${i}"
+        MISSING_INTERFACE="${i##*/}"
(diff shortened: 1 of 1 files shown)


ZeroChaos-:
This is a good start, but unfortunately it entirely breaks the function
```
+ iw phy /sys/class/ieee80211/phy0 interface add wlan0mon type monitor
+ IW_ERROR='command failed: No such file or directory (-2)' 
+ '[' -z 'command failed: No such file or directory (-2)' ']'
+ printf '\n\n ERROR adding %s mode interface: %s\n' monitor 'command failed: No such file or directory (-2)'
```

Also, please remove the two comments above your change since the comments are suggesting to do exactly what you are trying to do :-)

ZeroChaos- (changes requested):
Please fix setting MISSING_INTERFACE to the expected value.  You changed the meaning of ${i} but didn't update it's use everywhere

wileyhy:
You're right; thank you. Two fixes have commits up for review.

ZeroChaos- (approved):
tested on zsh, dash, bash, and mksh

Looks great, thanks!

ZeroChaos-:
merging, one unrelated check failure

ZeroChaos-:
thanks @wileyhy

wileyhy:
You're welcome! Thank _you_! :-)

Issue #2233, opened 2021-04-22
Aircrack-ng displays incorrect Transient Key


tinyboxvk:
# Issue type

- Defect - Incorrect value displayed

# System information

- OS: `Linux kali 5.10.0-kali7-amd64 #1 SMP Debian 5.10.28-1kali1 (2021-04-12) x86_64 GNU/Linux`
- CPU: Ryzen 3700X

# Aircrack-ng version

- Version: `1:1.6+git20210130.91820bc-1`

# Defect
## How to reproduce the issue

Incorrect PTK was shown when cracking using rockyou.txt. The key found was correct.

```
┌──(kali㉿kali)-[~]
└─$ aircrack-ng de-auth.cap -w /usr/share/wordlists/rockyou.txt
```

```
                               Aircrack-ng 1.6 

      [00:00:05] 104903/14344392 keys tested (23092.57 k/s) 

      Time left: 10 minutes, 16 seconds                          0.73%

                           KEY FOUND! [ P@$$w0rd ]


      Master Key     : D6 38 12 A0 F8 6B 11 EF C8 83 31 02 9C 8B AE B8 
                       7E FA 0B DE 2D D0 30 5C 27 90 BB FB F7 54 36 6D 

      Transient Key  : 3E B0 3C B6 EA 4D 91 E2 0D 06 52 9D 30 C9 9F 53 
                       56 EA 1B F7 6A 47 5C 07 17 55 88 00 00 00 00 00 
                       00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
                       00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 

      EAPOL HMAC     : 58 4C 33 E0 52 EB 53 DD 19 00 1A 33 9B 42 EE FE
```

Correct PTK was shown when cracking using a password file containing only the key.

```
┌──(kali㉿kali)-[~]
└─$ cat password.txt                             
P@$$w0rd
┌──(kali㉿kali)-[~]
└─$ aircrack-ng de-auth.cap -w password.txt
```

```
                               Aircrack-ng 1.6 

      [00:00:00] 1/1 keys tested (29.65 k/s) 

      Time left: --

                           KEY FOUND! [ P@$$w0rd ]


      Master Key     : D6 38 12 A0 F8 6B 11 EF C8 83 31 02 9C 8B AE B8 
                       7E FA 0B DE 2D D0 30 5C 27 90 BB FB F7 54 36 6D 

[...]

Issue #2553, opened 2023-07-12
PMK/TK/MIC output wrong in multiple thread mode


cheng3100:
# Defect Report

## Issue type

- [ ] Defect - Compilation or `make check` issue - Attach `config.log` and provide relevant system information such as `lscpu` - Make sure dependencies are installed
- [ ] Defect - Crash
- [x] Defect - Incorrect value displayed/received/stored
- [x] Defect - Unexpected behavior (obvious or confirmed in the GitHub Discussions)

## System information

- OS Name:  Ubuntu 22.04.2 LTS x86_64
- OS Version:
- Kernel version:  5.19.0-42-generic
- CPU: 
- Wireless card and chipset (if relevant):

## Aircrack-ng version

- Version:  Aircrack-ng 1.7 rev (The current mainline build version)
- Commit Revision hash:  d4e7705e 

### Details
When use the `aircrack-ng `'s passwd file parameter to find the passphrase with the pcap file. It will printout the current passphrase and PMK/TK/MIC information of one thread.  And it could happen that this routine print has a large chance to overwrite the finally right passphrase's PMK/TK/MIC.   

### How to reproduce the issue
Just enter the repo and use the test/ folder's file:
`aircrack-ng test/wpa2.eapol.cap -w test/password.lst`
![Animation](https://github.com/aircrack-ng/aircrack-ng/assets/25582840/fa44addc-a97b-42a7-8508-dd2888a4af0c)

you will see the passphrase is right but the PMK/TK/MIC is random calculate from one of password dictionary.

cheng3100:
I have submit a fix here: #2554

PR #2554, opened 2023-07-12
Fix PMK output wrong due to multiple thread overwrite.

cheng3100:
When one of threads find the right passphrase and print it out. The other thread may still have some passphrase list to run and could finally overwrite the right PMK output. #2553


tinyboxvk:
Might also fix https://github.com/aircrack-ng/aircrack-ng/issues/2233

cheng3100 (commented):
show_wpa_stats is the error point which will overwrite the correct finally output

Issue #2556, opened 2023-07-20
airmon-ng: lines 134 & 137, `[ "$(find ... | wc -l)" != '0' ]` will always return true


wileyhy:
# Defect Report

## Issue type

- [y] Defect - Incorrect value displayed/received/stored

## System information

[liveuser@localhost-live]$ `cat /etc/os-release`

> NAME="Fedora Linux"
> VERSION="38 (Xfce)"
> ID=fedora
> VERSION_ID=38
> VERSION_CODENAME=""
> PLATFORM_ID="platform:f38"
> PRETTY_NAME="Fedora Linux 38 (Xfce)"
> ANSI_COLOR="0;38;2;60;110;180"
> LOGO=fedora-logo-icon
> CPE_NAME="cpe:/o:fedoraproject:fedora:38"
> DEFAULT_HOSTNAME="fedora"
> HOME_URL="https://fedoraproject.org/"
> DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f38/system-administrators-guide/"
> SUPPORT_URL="https://ask.fedoraproject.org/"
> BUG_REPORT_URL="https://bugzilla.redhat.com/"
> REDHAT_BUGZILLA_PRODUCT="Fedora"
> REDHAT_BUGZILLA_PRODUCT_VERSION=38
> REDHAT_SUPPORT_PRODUCT="Fedora"
> REDHAT_SUPPORT_PRODUCT_VERSION=38
> SUPPORT_END=2024-05-14
> VARIANT="Xfce"
> VARIANT_ID=xfce
> 

[liveuser@localhost-live]$ `uname -a`

> Linux localhost-live 6.3.7-200.fc38.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Jun  9 15:21:11 UTC 2023 x86_64 GNU/Linux

[liveuser@localhost-live]$ `lscpu`

> Architecture:            x86_64
>   CPU op-mode(s):        32-bit, 64-bit
>   Address sizes:         39 bits physical, 48 bits virtual
>   Byte Order:            Little Endian
> CPU(s):                  8
>   On-line CPU(s) list:   0-7
> Vendor ID:               GenuineIntel
>   Model name:            Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz
>     CPU family:          6
>     Model:               142
>     Thread(s) per core:  2
>     Core(s) per socket:  4
>     Socket(s):           1
>     Stepping:            10
[...]

PR #2557, opened 2023-07-20
airmon-ng.linux: add '-type l' to two `find` command; fixes bug #2556

wileyhy:
The search directory was being included in find's results.


ZeroChaos-:
nice catch, thanks!

Issue #2558, opened 2023-07-30
airmon-ng: line 263, in a for loop, replace ls with a glob


wileyhy:
# Defect Report

## Issue type

- [x] Defect - Unexpected behavior (obvious or confirmed in the GitHub Discussions)

## System information

```sh
$ cat /etc/os-release
NAME="Fedora Linux"
VERSION="38 (Xfce)"
ID=fedora
VERSION_ID=38
VERSION_CODENAME=""
PLATFORM_ID="platform:f38"
PRETTY_NAME="Fedora Linux 38 (Xfce)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:38"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f38/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=38
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=38
SUPPORT_END=2024-05-14
VARIANT="Xfce"
VARIANT_ID=xfce
 $ uname -a
Linux localhost-live 6.3.7-200.fc38.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Jun  9 15:21:11 UTC 2023 x86_64 GNU/Linux
 $ lscpu
Architecture:            x86_64
  CPU op-mode(s):        32-bit, 64-bit
  Address sizes:         39 bits physical, 48 bits virtual
  Byte Order:            Little Endian
CPU(s):                  8
  On-line CPU(s) list:   0-7
Vendor ID:               GenuineIntel
  Model name:            Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz
    CPU family:          6
    Model:               142
    Thread(s) per core:  2
    Core(s) per socket:  4
    Socket(s):           1
    Stepping:            10
    CPU(s) scaling MHz:  84%
    CPU max MHz:         3400.0000
    CPU min MHz:         400.0000
    BogoMIPS:            3600.00
[...]

PR #2559, opened 2023-07-30
airmon-ng.linux: fix shellcheck error SC2045 at lines 261-264, 

wileyhy:
line 261, Remove explanatory comment.
line 262, Remove "shellcheck disable" comment.
line 263 > 261, Change an unquoted command substitution containing an `ls` command into a glob expression on the same directory path.  
line 264 > 261, Move the "type" directory name up one line. 
line 262, Add a test `[` for the existence of such a directory ( "$j" ). 
line 264 > 262, Change the argument to the `cat` command. 
lines 263 & 264, Add two parameter expansions of "$j". 
Keep the total line count and the later variables "$k" and "$i" on lines 266 and 267, respectively, the same as they were.


wileyhy:
Fixes bug #2558

wileyhy:
Hi, 
  Line 1331 has a comment: "# refactor this for globs, but make sure it works across shells," referring to the same shellcheck error, SC2045.  Which shells should I test for (and / or versions)?  A comment in #2551 mentions "zsh, dash, bash, and mksh."

PR #2580, opened 2023-09-30
autotools: reset PCRE CFLAGS/LIBS with both PCRE and PCRE2 present

Ansuel:
Commit b381ef3f6b6c ("autotools: indicate if PCRE or PCRE2 is being used") fixed a case where both pcre and pcre2 library are detected and put a preference on using pcre2.

Although the commit fix this corner case, there is still a latent problem with trying to link/include both library. This is caused by the fact that in the Makefile.inc for src and lib, we include both PCRE_CFLAGS and PCRE2_CFLAGS and PCRE_LIBS and PCRE2_LIBS for each tool/lib.

To handle this and not bloat the Makefile with additional condition, simply reset the PCRE_CFLAGS and PCRE_LIBS in case where we detect both library and we prefer to use pcre2.

Fixes: b381ef3f6b6c ("autotools: indicate if PCRE or PCRE2 is being used")


Mister-X-:
Could you explain the use case of providing both to `./configure`?

Ansuel:
No it's not about providing both but it's about having both present in the system where this is built... The same logic is also used in other project where both pcre are detected and pcre2 have priority

Issue #2585, opened 2023-10-19
Unable to quit using CTRL+C


garsaud:
# Defect Report

## Issue type

- [ ] Defect - Compilation or `make check` issue - Attach `config.log` and provide relevant system information such as `lscpu` - Make sure dependencies are installed
- [x] Defect - Crash
- [ ] Defect - Incorrect value displayed/received/stored
- [x] Defect - Unexpected behavior (obvious or confirmed in the GitHub Discussions)

## System information

- OS Name: MacOS, Archlinux
- OS Version: MacOs 13.6, Arch 231001.0.182270
- Kernel version: Linux 385bac7603bc 5.15.49-linuxkit
- CPU: Dual-Core Intel Core i7 @ 3.5 GHz
- Wireless card and chipset (if relevant):

## Aircrack-ng version

- Version: 1.7
- Commit Revision hash: e82138cb

### Details

Using the non-quiet mode, when choosing the target network, it is not possible to quit the program using CTRL+C. The only way I found to do it "properly" is by selecting "0" (or a non-existant target). If CTRL+C is pressed, “Quitting aircrack-ng...” is displayed, the process crashes, and you’ll have to kill it manually.

![Screenshot of Aircrack prompting for a target, and user trying to quit desperately](https://github.com/aircrack-ng/aircrack-ng/assets/3667366/dc2a3dad-4d46-4ceb-9293-460686de399c)

### How to reproduce the issue

Having a dictionnary and a capture file with targets inside, launch `aircrack-ng -w dictionnary.txt capture.pcap`, then try to quit instead of selecting a target.

aps221292:
hi

PR #2586, opened 2023-10-19
Fix quitting aircrack-ng using CTRL+C

garsaud:
Fixes issue #2585

Allow for a user to quit while Aircrack prompts for a target.

Result:
Screenshot of aircrack successfully closed using ctrl+c


Mister-X-:
Now it coredumps when using Ctrl-C while running.

PR #2592, opened 2023-11-27
Improve, simplify shell script

h3xx:
- Fix unbound shell variable warning from `set -u` about `GITHUB_SHA` being unbound. Seen when running build locally.
- Remove obsolete SVN support
  - This section had a bug in it anyway -- `$REVISION` could never be empty since it starts out with `REVISION="_r$(...)"` - it would always start with `_r` regardless of what the subshell echoed, so yeah, never empty.
  - `svn.aircrack-ng.org`'s DNS record is also no longer up
- Fix various quoting issues reported by shellcheck.
- Avoid pedantic variable checking by just using defaults (`${VAR:-'default value'}`)
- Change else-if chain to `case`

### DIFF
--- a/evalrev
+++ b/evalrev
@@ -2,17 +2,12 @@
 
 set -euf
 
-DIR=${1:-}
+DIR=${1:-'.'}
 CMD=${2:-}
 
-SCM=""
+SCM=
 
-if [ x$DIR = "x" ]
-then
-	DIR="."
-fi
-
-if test -d "${DIR}/.git"; then
+if [ -d "$DIR/.git" ]; then
     #
     # Locate the closest annotated tag
     #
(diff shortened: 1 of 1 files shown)


h3xx:
Oh, and if you prefer I *fix* the SVN support rather than remove it, I can do that too.

evalrev
```diff
@@ -2,17 +2,12 @@
...
 
-DIR=${1:-}
+DIR=${1:-'.'}
 CMD=${2:-}
 
-SCM=""
+SCM=
 
-if [ x$DIR = "x" ]
-then
```
Tusenka, line 11: why we could remove this?

evalrev
```diff
@@ -28,40 +23,24 @@ if test -d "${DIR}/.git"; then
...
-        REVISION="_r1"
-        SCM="_r1"
-    fi
-elif [ -n "${GITHUB_SHA}" ]; then
-    SHORT_SHA_REVISION="$(echo ${GITHUB_SHA} | cut -b 1-8)"
+elif [ -n "${GITHUB_SHA:-}" ]; then
+    SHORT_SHA_REVISION="$(echo "$GITHUB_SHA" | cut -b 1-8)"
 
-    VERSION=$(cat ${DIR}/AC_VERSION 2>/dev/null)
+    VERSION=$(cat "$DIR/AC_VERSION" 2>/dev/null)
```
Tusenka, line 29: why removed {}?

evalrev
```diff
@@ -28,40 +23,24 @@ if test -d "${DIR}/.git"; then
...
 
-if test "x$CMD" = "xscm"; then
-    echo $SCM
-elif test "x$CMD" = "xquoted"; then
-    echo \"$REVISION\"
-else
-    echo $REVISION
-fi
+case $CMD in
+    'scm')
```
Tusenka, line 37: awesome!

Tusenka (approved):

evalrev
h3xx: I removed it because, despite what shellcheck may tell you, the {} is not necessary, it's just extra characters in the line.

I have a hypothesis that whoever wrote this originally thought that `${VAR}` was the same thing as `"$VAR"` for the purposes of quoting. It's an easy mistake to make.

Tusenka: I see
https://www.gnu.org/software/bash/manual/html_node/Word-Splitting.html

Tusenka (approved):

PR #2615, opened 2024-06-11
Airodump GPSd 3.20 fix

TonyCalzoni:
Changes located in /src/airodump-ng/airodump-ng.c, lines 4776 and 4777
Removed semicolon from within ?WATCH command string on line 4776, and decreased the corresponding length value of 22 to 21

### DIFF
--- a/include/aircrack-ng/support/communications.h
+++ b/include/aircrack-ng/support/communications.h
@@ -142,6 +142,7 @@ struct communication_options
 
 	int f_index; /* outfiles index       */
 	FILE * f_txt; /* output csv file      */
+	FILE * f_json; /* output json file      */    
 	FILE * f_kis; /* output kismet csv file      */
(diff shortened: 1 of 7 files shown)


Mister-X-:
This PR has not only GPSd but some JSON things in there. They should be separate. I'll also need an explanation on why removing the semicolon fixes GPSd in the PR.

TonyCalzoni:
That's the original code I pulled from the repo. It fixes it because the
GPSd server does not like the semicolon at the end of the watch command.
Json is how GPSd likes to talk to your client, airodump, which again is
your original code.

The json dump handling I pulled into my repo for testing.
All I care about is documented protocol noncompliance per GPSd API on line 4776 in airodump-ng.c and the associated length check on line 4777.
As of right now, I'm the only person with an up to date build of aircrack-ng that works with up to date gpsd.
This is something that should have been found in testing, but wasn't.

PR #2625, opened 2024-08-10
Updated README

TakangNina:


README
```diff
@@ -68,14 +68,14 @@ Aircrack-ng
 
 	Problem: On windows only, opening/creating a database doesn't work when airolib-ng is in directories containing
 	         special characters like 'ç', 'é', 'è', 'à', ... (directories containing spaces are not affected).
-	Reason: It's a SQLite issue.
+	Reason: It's an SQLite issue.
```
Mister-X-, line 71: For future reference, pronunciation varies by vendor, and in the case of SQLite, this link details the answer: https://english.stackexchange.com/questions/431329/what-is-the-correct-pronunciation-of-sqlite

PR #2630, opened 2024-09-26
Replace potentially insecure string functions with their more secure counterparts

gresskar:
Bugs reported by the clang static analyzer


ZeroChaos- (approved):
This looks good to me but I will let someone who knows C hit merge :-)

PR #2667, opened 2025-06-19
airmon-ng: workaround multiple phy indices in rfkill

liovbt:
While moving wiphys in and out network namespaces, I noticed that rfkill can get confused and list the same phy multiple times.

$ rfkill list
6: phy5: Wireless LAN
        Soft blocked: no
        Hard blocked: no
7: phy5: Wireless LAN
        Soft blocked: no
        Hard blocked: no

This change attempts to use the first index instead of failing due to invalid parsing of the output.

### DIFF
--- a/scripts/airmon-ng.linux
+++ b/scripts/airmon-ng.linux
@@ -307,7 +307,7 @@ rfkill_check() {
 	fi
 	#first we have to find the rfkill index
 	#this is available as /sys/class/net/wlan0/phy80211/rfkill## but that's a bit difficult to parse
-	index="$(rfkill list | grep "${1}:" | awk -F: '{print $1}')"
+	index="$(rfkill list | grep "${1}:" | awk -F: '{print $1}' | head -n 1)"
 	if [ -z "$index" ]; then
 		return 187
 	fi


ZeroChaos-:
I don't have a problem with the intended change, but the unintended change will break function entirely for multi-phy systems.  Please fix the grep "${2}" back to "${1}"

liovbt:
Oops sorry for the unintended glitch. I have fixed it.

ZeroChaos-:
It happens.  I would love to know how you got the kernel that confused that you hit this issue though :-)

liovbt:
I am running Kali Linux on a RPi 3 and I have a WiFi adapter AWUS036NHA (Atheros AR9271), using driver ath9k_htc. I had the problem a few times already. Here is an excerpt of what I did today, but I may have done other things in the middle with other adapters connected to the RPi. Not sure if systematically reproduced.
```
# iw phy phy5 set netns name tmp
# ip netns exec tmp bash
# airmon-ng start wlu1u3 1
rfkill error: rfkill: invalid identifier: '6
7'
```

ZeroChaos-:
Thanks for the fix

PR #2686, opened 2025-10-23
draft pr

Rishabh0712:


RealEnder:
Using --anonymize_packet option will damage the authentication sequence, so the extracted hashes will be uncrackable for sure. Please at least add a message to the users about this consequence in the help.

ZerBea:
What exactly is the purpose of this option?
If the user decide to use an online service or to share the dump file, he must attach the mapping file. Now nothing is anonymized anymore. If he doesn't attach it, the authentication sequences are destroyed. 

If you decide, to merge the PR, please consider to add an information directly to the dump file that it's no longer an original dump file and that the MAC ADDRESSES have been modified.
Since the cap/pcap dump file format doesn't allow to add comments, I suggest to move to pcapng dump file format. Now it is easier for other tools to identify the origin of the dump file as (aircrack-ng suite):
```
shb_userappl:
    The shb_userappl option is a UTF-8 string containing the name of the application used to create this section. The string is not zero-terminated.
```
https://www.ietf.org/archive/id/draft-tuexen-opsawg-pcapng-03.html#name-section-header-block

And please add s comment that it was subsequently changed (e.g. by custom option):
https://www.ietf.org/archive/id/draft-tuexen-opsawg-pcapng-03.html#name-custom-options

BTW:
ESSID is missing in the example dump file (anonymized-wpa.pcap). That makes the 4way handshake useless.

Issue #402, opened 2018-03-10
zd1211rw injection patch needs updating


aircrack-ng:
**Reported by maleadt on 18 Mar 2008 09:00 UTC**

The current patch for the Zydax 1211rw driver, which has to enable monitoring and injection in the driver, does not work anymore on 2.6.24. As the zd1211rw driver has been modified in the 2.6.24 release (monitoring support has been added now), I can't seem to properly modify the patch (some hunks fail in zd_mac.c, and I guess some patching has been made redundant by the native monitoring support)...

aircrack-ng:
**Modified by misterx on 18 Mar 2008 09:00 UTC**

aircrack-ng:
**Comment by wicher on 18 Mar 2008 09:00 UTC**

Patch from http://www.zlaten.biz/tmp/zd1211rw-inject+dbi-fix-2.6.26.patch applies nicely to linux-2.6.37-git2 and works fine .

aircrack-ng:
**Comment by misterx on 18 Mar 2008 09:00 UTC**

Committed in aircrack-ng/aircrack-ng@1ef19eda601f362b92a820ce01dbb039e1bc4e1a.

aircrack-ng:
**Comment by misterx on 18 Mar 2008 09:00 UTC**

Milestone N/A deleted

PR #2702, opened 2026-02-22
Introduce native nl80211 support for channel and frequency management with WEXT fallback

fjh1997:
Modern Linux distributions are beginning to deprecate and remove support for Wireless Extensions (WEXT). For example, Fedora has recently disabled the `CONFIG_CFG80211_WEXT` kernel option by default (see [[Fedora Changes: RemoveWirelessExtensions](https://fedoraproject.org/wiki/Changes/RemoveWirelessExtensions)](https://fedoraproject.org/wiki/Changes/RemoveWirelessExtensions)).

Because `aircrack-ng` (specifically tools like `aireplay-ng`) previously relied on `ioctl` (WEXT) to determine and set channels, running it on these newer kernels results in a failure with the following error message:

> "Couldn't determine current channel for wlan0mon, you should either force the operation with --ignore-negative-one or apply a kernel patch."
> *(Reference: [[morrownr/USB-WiFi issue #402](https://github.com/morrownr/USB-WiFi/issues/402#issuecomment-3940093906)])*

This PR resolves the issue by migrating the core channel and frequency management functions to the modern `nl80211` Netlink API, ensuring seamless compatibility with modern Linux kernels.

## Proposed Changes

* **Implemented Netlink API functions:** Added `linux_get_channel_nl80211`, `linux_get_freq_nl80211`, and `linux_set_freq_nl80211` to handle frequency and channel operations natively via `nl80211`.
* **Added a Netlink callback:** Introduced `get_freq_nl_callback` to accurately parse `NL80211_ATTR_WIPHY_FREQ` from the Netlink message payload.
* **Robust Fallback Mechanism:** Designed all new `nl80211` functions to gracefully fall back to the legacy `ioctl` (WEXT) implementations (`linux_get_channel`, `linux_get_freq`, etc.) if the Netlink interface fails or is unsupported by the driver.

## Testing Instructions

1. Compile `aircrack-ng` on a system with a modern kernel where `CONFIG_CFG80211_WEXT` is disabled (e.g., Fedora 43).
2. Run `aireplay-ng` against a monitor mode interface (e.g., `wlan0mon`).
3. Verify that the tool can successfully determine and set the channel without throwing the "Couldn't determine current channel" error.
4. Test on an older kernel/driver to verify that the `ioctl` fallback works correctly without regressions.

### DIFF
--- a/lib/osdep/linux.c
+++ b/lib/osdep/linux.c
@@ -363,6 +363,20 @@ static void nl80211_cleanup(struct nl80211_state * state)
 
 /* Callbacks */
 
+/* --- ADDED NETLINK CALLBACK FOR GETTING FREQUENCY --- */
+static int get_freq_nl_callback(struct nl_msg *msg, void *arg)
+{
+	struct genlmsghdr *gnlh = nlmsg_data(nlmsg_hdr(msg));
+	struct nlattr *tb[NL80211_ATTR_MAX + 1];
+	int *freq = arg;
+
+	nla_parse(tb, NL80211_ATTR_MAX, genlmsg_attrdata(gnlh, 0), genlmsg_attrlen(gnlh, 0), NULL);
+	
+	if (tb[NL80211_ATTR_WIPHY_FREQ]) {
+		*freq = nla_get_u32(tb[NL80211_ATTR_WIPHY_FREQ]);
+	}
+	return NL_SKIP;
+}
 /*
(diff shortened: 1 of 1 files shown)


gianfelicevincenzo:
I hope this PR will be implemented as soon as possible because **_currently aircrack on fedora (more generally on new kernels) does not work_**

kimocoder:
WIFI 7 requires NETLINK support :+1:

king34992 (commented):
Ok

Issue #2516, opened 2023-03-13
Fix shellcheck issues in airmon-ng


Mister-X-:
# Description

There are still a few things to fix in airmon-ng, that were reported by shellcheck. They are ignored for the time being so that linting (#2501) won't fail.

This is just to keep track of them so they're not forgotten in the code. There are approximately 20 of them (Search for `disable=`).

wileyhy:
One of these was fixed with #2551

wileyhy:
Here's a current list as of today. 

```sh
 $ grep -oEe 'disable=(SC)?[0-9]{4}' ./scripts/airmon-ng.linux | sort | uniq -c
      8 disable=2009
      1 disable=2012
      4 disable=2045
      2 disable=2059
      1 disable=2166
      2 disable=2181
```

ZeroChaos-:
As noted in the comments, some of those are correct and NOT going to be removed due to compatibility requirements.  Just grepping for disable is not providing an accurate list.  Off the top of my head all the warnings about ps simply don't apply to us.

wileyhy:
Okay.  ...that's true, some systems still don't have `pgrep` yet

ZeroChaos-:
It's mostly that isn't not default and I didn't want to force the install.  The warning is right, ps is harder to use, but it does work :-)

PR #2711, opened 2026-06-08
airmon-ng: fix shellcheck warnings SC2059 and SC2181

BLABLA1000gg:
## Summary

Fixes two remaining shellcheck warnings from #2516:

- **SC2059**: `printf` calls were using shell variables directly as the format string. Replaced with explicit format specifiers (`%s`, `\t`, `\n`). This avoids potential issues when variables contain `%` or `\` sequences and satisfies shellcheck.

- **SC2181**: Two instances of `if [ $? != 0 ]` were checking the exit code after a command substitution assignment. Replaced with `if ! cmd; then` pattern, which is idiomatic POSIX sh and avoids the pitfall of $? being clobbered between the command and the check.

## Changes

- `scripts/airmon-ng.linux`: 3 suppressions removed, equivalent logic preserved

## Testing

The output format of both `printf` statements is identical — only the quoting style changes from variable-as-format to positional arguments. The rfkill error handling logic is functionally unchanged.

### DIFF
--- a/scripts/airmon-ng.linux
+++ b/scripts/airmon-ng.linux
@@ -311,10 +311,7 @@ rfkill_check() {
 	if [ -z "$index" ]; then
 		return 187
 	fi
-	rfkill_status="$(rfkill list "${index}" 2>&1)"
-	# it's easier to capture and check this way, but a refactor could make it simpler and safer
-	# shellcheck disable=2181
-	if [ $? != 0 ]; then
+	if ! rfkill_status="$(rfkill list "${index}" 2>&1)"; then
 		printf "rfkill error: %s\n" "${rfkill_status}"
 		return 187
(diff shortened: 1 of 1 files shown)


ZeroChaos- (approved):

ZeroChaos-:
This looks good enough to me, thanks!

ZeroChaos-:
CI failures unrelated, merging.

ZeroChaos-:
Oops, I should have tested before merging

```
zero@naga aircrack-ng % ./scripts/airmon-ng.linux                                                                                                                             (git)-[master] 

PHY	Interface	Driver		Chipset

phy0	wlan0\t\tiwlwifi\t\tIntel Corporation 700 Series Chipset CNVi WiFi (rev 11)
```

I had to revert this change, please feel free to reopen after it actually works.

PR #2717, opened 2026-06-30
airodump-ng: add `--mfp` / `--pmf` to display Management Frame Protection AP/Client configuration

r4ulcl:
This PR adds an optional column to the airodump-ng **live terminal display** showing the MFP / 802.11w (Protected Management Frames) state of each AP and station.

## What it does

- New flag `--mfp` (alias `--pmf`, since both names are common) enables an extra `MFP` column in both the AP and station tables.
- Reported value per network:
  - blank / `-1`: unknown (not yet seen)
  - `0`: disabled
  - `1`: capable
  - `2`: required
- Scope: display only. It only changes the on-screen table.
- The state is read from the RSN IE capabilities bits (bit 6 = required, bit 7 = capable) in beacons / probe responses for APs, and from the RSN IE in (re)association requests for stations.
- Open, WEP and WPA1 networks are reported as `0` (MFP only applies to RSN / WPA2-WPA3).
- A client's MFP value only appears once airodump captures that client's (re)association request, which carries the RSN IE.

## Changes

- Add `mfp` field to `struct AP_info` and `struct ST_info`, initialized to `-1`.
- Add `rsn_mfp_from_ie()` / `mfp_value_from_caps()` helpers to parse the RSN capabilities.
- Add the `--mfp`/`--pmf` long options, the `F` short option and the usage line.
- Render the new column in `dump_print()` and adjust column widths; minor table spacing / probes-width fixes.
- Wired the same way as the existing `--wps` flag (long/short option, optional column, per-AP/station rendering).

### DIFF
--- a/include/aircrack-ng/support/station.h
+++ b/include/aircrack-ng/support/station.h
@@ -179,6 +179,7 @@ struct AP_info
 
 	int marked;
 	int marked_color;
+	int mfp; /* -1 = unknown, 0 = not enabled, 1 = capable, 2 = required */
 	struct WPS_info wps;
 };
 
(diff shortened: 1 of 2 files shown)


satyamnetam353-lgtm:
https://github.com/WiFi password/aircrack-ng.git

Issue #2179, opened 2020-09-05
Airmon-ng: handle Qualcomm qcacld monitor mode


kimocoder:
Add support to Qualcomm monitor mode on internal chipset. 40% of mobile units uses their chipset.

**Start wlan0 in monitor mode:**
ip link set wlan0 down
echo 4 > /sys/module/wlan/parameters/con_mode;
ip link set wlan0 up

**Stop wlan0 monitor mode:**
ip link set wlan0 down
echo 0 > /sys/module/wlan/parameters/con_mode
ip link set wlan0 up

@Mister-X-

ZeroChaos-:
What is the driver detected as by airmon-ng?
Show me "airmon-ng --verbose'" and I can add this pretty easily.

kimocoder:
![IMG_20200909_012220_516](https://user-images.githubusercontent.com/4252297/92536885-122aab80-f23b-11ea-8838-c6aef8fb2619.jpg)

ZeroChaos-:
can I also see the output of this please?
```
iw phy phy0 info | grep -q 'interface combinations are not supported'
```

kimocoder:
That would be 'None'

ZeroChaos-:
sorry, I need the exit code.  That or just drop the -q and give me the output.

yesimxev:
None with or without -q @kimocoder @ZeroChaos-

kimocoder:
Thanks heaps 💯

ZeroChaos-:
Does channel control with iw work for this device?

ZeroChaos-:
I took a wild stab at this based on the information provided thus far.  Please let me know how it goes....

yesimxev:
> Does channel control with iw work for this device?

Unfortunately, not. Seems it's a lightweight qcacld-3.0 driver though https://github.com/yesimxev/android_kernel_nokia_sdm660/tree/lineage-17.1/drivers/staging/qcacld-3.0

ZerBea:
@yesimxev, does channel control work with ioctl() system calls (e.g.: using iwconfig) instead of NETLINK messages?

yesimxev:
No, all I could see is if Android isn't controlling (svc wifi is down?) Then it's 0

kimocoder:
@ZerBea Welcome to our party 🥇

ZerBea:
@yesimxev , thanks.
[...]

PR #2720, opened 2026-07-27
airmon-ng: find qcacld con_mode instead of assuming module name

iqzer0:
### Problem

The qcacld monitor mode support added in #2179 makes two assumptions that no longer hold on GKI-era devices:

1. `qcaldSafetyCheck()`, `changeQcacldIfaceTypeMonitor()` and `changeQcacldIfaceTypeManaged()` hardcode `/sys/module/wlan/parameters/con_mode`.
2. `startMac80211Iface()` and `stopMac80211Iface()` match `DRIVER = "icnss"` exactly.

Both are correct on the kernel 4.x devices that support was developed against, where qcacld builds as a generic `wlan.ko`. On GKI devices qcacld is built per target, so the module carries a target specific name and the interface reports a suffixed driver. On a Redmi Note 13 Pro 5G (SM7435, kernel 5.10 GKI, LineageOS 23.2):

```
$ ls /sys/module/*/parameters/con_mode
/sys/module/qca_cld3_adrastea/parameters/con_mode

$ awk -F= '$1=="DRIVER"{print $2}' /sys/class/net/wlan0/device/uevent
icnss2
```

So `qcaldSafetyCheck()` aborts with `Unable to find ... which is needed to control mode`, the qcacld branches are never taken, and monitor mode fails:

```
# airmon-ng start wlan0
ERROR adding monitor mode interface: command failed: Invalid argument (-22)
```

### Change

- Add `qcacldConModePath()`, which locates `con_mode` by globbing `/sys/module/*/parameters/con_mode` and caches the result in `QCACLD_CON_MODE`. Used by the safety check and by both mode writes.
- Match the `icnss` prefix instead of the exact string, so `icnss2` (and any future suffix) is handled.

This keeps the existing behaviour on devices where the module is `wlan` and the driver is `icnss`, and additionally covers the newer naming. It also removes the need to update the script again the next time a target specific module name appears.

The sequence itself is unchanged: `link down` -> `con_mode` -> `link up` -> set channel.

### Testing

On the device above, before the change:

```
# airmon-ng start wlan0
ERROR adding monitor mode interface: command failed: Invalid argument (-22)
# echo $?
1
```

After:

```
# airmon-ng start wlan0
                (monitor mode enabled)
# cat /sys/module/qca_cld3_adrastea/parameters/con_mode
4
# cat /sys/class/net/wlan0/type
803                                   # ARPHRD_IEEE80211_RADIOTAP
# iw dev
        Interface wlan0
                type monitor
```

Capture works (BSSIDs redacted):

```
# tcpdump -i wlan0 -c 5 -e -nn -Z root
... 1.0 Mb/s 2457 MHz 11b -74dBm signal -96dBm noise antenna 0 Beacon ... ESS CH: 10, PRIVACY
... 18.0 Mb/s 2457 MHz 11g -64dBm signal ... Unhandled Management subtype(7)
... 18.0 Mb/s 2457 MHz 11g -66dBm signal ... Authentication
5 packets captured
33 packets received by filter
0 packets dropped by kernel
```

And teardown:

```
# airmon-ng stop wlan0
                (monitor mode disabled)
# cat /sys/module/qca_cld3_adrastea/parameters/con_mode
0
# cat /sys/class/net/wlan0/type
1
```

Exactly one module exposes `con_mode` on this device, so the glob is unambiguous. I do not have a `wlan.ko` era device to hand, so a check on one of those would be welcome.

### Out of scope

This only fixes monitor mode enablement. `aireplay-ng` injection is a separate, driver side matter on this hardware: every wireless extensions ioctl returns `EOPNOTSUPP` (verified for `SIOCGIWMODE`, `SIOCSIWMODE`, `SIOCSIWFREQ`, `SIOCSIWRATE`), so `set_monitor()` in `lib/osdep/linux.c` cannot succeed regardless of this change. I have deliberately kept that out of this PR.

### DIFF
--- a/scripts/airmon-ng.linux
+++ b/scripts/airmon-ng.linux
@@ -430,10 +430,27 @@ changeMac80211IfaceTypeMonitor() {
   printf "\t\t(monitor mode enabled)\n"
 }
 
+qcacldConModePath() {
+  # The qcacld module is not always called "wlan". On GKI-era devices it is
+  # built per-target, e.g. qca_cld3_adrastea, so locate con_mode instead of
+  # assuming the module name.
+  if [ -n "${QCACLD_CON_MODE}" ]; then
+    printf "%s" "${QCACLD_CON_MODE}"
+    return 0
+  fi
+  for f in /sys/module/*/parameters/con_mode; do
+    [ -f "${f}" ] || continue
+    QCACLD_CON_MODE="${f}"
+    printf "%s" "${QCACLD_CON_MODE}"
+    return 0
+  done
+  return 1
+}
+
 qcaldSafetyCheck() {
-  icnss_file="/sys/module/wlan/parameters/con_mode"
-  if [ ! -f "${icnss_file}" ]; then
-    printf "Unable to find %s which is needed to control mode.\n" "${icnss_file}"
+  icnss_file="$(qcacldConModePath)"
+  if [ -z "${icnss_file}" ] || [ ! -f "${icnss_file}" ]; then
+    printf "Unable to find a con_mode parameter under /sys/module/*/parameters/\n"
     printf "Is /sys mounted correctly?\n"
     exit 1
   fi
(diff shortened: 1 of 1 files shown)


kimocoder:
I will make proper support for QCACLD-3.0 in a separate PR; got full overview of both the devices, QDF_MONITOR_MODE and the packet (frame) injection for the QCACLD :+1: 

I added a IOCTL, + other patches I've got laying around too if needed, over at #2723

PR #2723, opened 2026-08-04
Patch set; various minor fixes + arm/aarch64 env ++

kimocoder:
# Patch sets collection

Seven commits, `4c41bc71..cb33bffa`. Six files, +245/−69. Every one builds
warning-free and passes `make check` (48/48).

---

### Command injection through the interface name (CWE-78)

`lib/osdep/linux.c` built roughly nine shell command lines out of the
user-supplied interface name and ran them through `system()`. The only guard
was a length check against `IFNAMSIZ`, so any name under 16 characters
containing shell metacharacters was executed — and these tools run as root:

```console
$ airodump-ng 'a;id>/tmp/pw #'
$ cat /tmp/pw
uid=0(root) gid=0(root) ...
```

This was reachable from every tool that opens an interface through `wi_open()`:
airodump-ng, aireplay-ng, airtun-ng, besside-ng, tkiptun-ng, wpaclean and
airbase-ng. A second tainted path existed in the ipw2200 branch, where the rtap
interface name read back from sysfs was used verbatim, trailing newline
included.

The shell was removed from these paths rather than escaped:

- `is_valid_iface_name()` accepts only what the kernel itself accepts in an
  interface name, and is applied in `do_linux_open()`, `opensysfs()`,
  `is_ndiswrapper()` and to the name read from sysfs.
- The three `ls /sys/class/net/%s/...` probes became `stat()` calls.
- The `wlancfg`/`iwpriv` grep pipelines became `program_output_contains()` —
  fork, pipe, exec, matching in C, carrying a prefix between reads so a match
  straddling a read boundary is not missed.
- `sysctl`, `iwpriv rfmontx` and `ifconfig up` became `run_program()`, a plain
  fork/execvp.

No `system()` call handling interface data remains in the file.

*Commit `f66e9557` — CodeQL alert 68, critical.*

---

## Undefined behaviour from unchecked `scanf` results

Three instances of the same mistake: a `scanf`-family result compared only
against zero, when the function also returns `EOF`.

### Hex wordlist parsing

`next_key()` in `src/aircrack-ng/aircrack-ng.c` checked
`sscanf(hex, "%x", &dec) == 0`. A whitespace-only token returns `EOF`, not `0`,
so it slipped through and `dec` was read having never been assigned. A
malformed wordlist line therefore produced a wrong candidate key instead of
being skipped. Now compared against the expected conversion count.

*Commit `5ba35a92` — alert 74.*

### Battery state

`get_battery_state()` in `lib/osdep/linux.c` asked for five conversions but
only rejected zero, so a short or malformed `/proc/apm` line was accepted part
way through and `ac`, `charging`, `flag` and `units` were used uninitialised.
Now requires all five.

*Commit `aa639294` — alert 73.*

### Memory size

`get_ram_size()` in `lib/libac/support/common.c` looped while `fscanf` returned
non-zero. `EOF` is non-zero, so on end of file the body ran an extra pass
against stale values, and on a read error — where `feof()` stays clear — the
loop never ended.

Worth noting: the textbook fix of comparing against the conversion count would
have broken this loop. `/proc/meminfo` values carry a `kB` suffix, which is read
as the next name, making `%d` fail and return 1; that partial match has to keep
the loop alive so the following pass can realign. The loop now stops on `EOF`
instead, which preserves every existing result.

*Commit `a75d49a1` — alert 72.*

---

## airmon-ng on phys without a bus device

A phy created by a virtual driver has no `/sys/class/ieee80211/<phy>/device`
symlink, so every check written in terms of `device/net` misreads it. On a
Kalidroid VM the Android wiphy is exactly that, and it lists `managed` as its
only interface mode, so airmon-ng offered to assign an interface on every
single run — including `airmon-ng --version` — and then failed with
`Operation not supported (-95)`.

Four defects, all in that area:

- Lost-phy detection tested `device/net`, so any virtual phy read as having no
  interfaces regardless of what was attached to it.
- Monitor support was never checked before attempting the attach, so on a phy
  that cannot do monitor the only possible outcome was `-95`.
- The post-creation check used the same path, so a successful
  `iw interface add` on such a phy would report
  `Unable to create ... and no error received`.
- `yesorno()` recursed on any unrecognised input, including the empty read it
  gets when stdin is closed. Running airmon-ng from a script or pipe spun until
  the shell gave up — 999 "Invalid input" lines before it died.

Fixed with `phyInterfaces()`, which resolves a phy's interfaces through the
`phy80211` link every wireless netdev carries and works for virtual and
bus-backed phys alike; `phySupportsMonitor()`, which skips phys that cannot
hold a monitor interface; and treating a closed stdin as "no".

*Commit `cb33bffa`.*

---

### Flaky airdecap-ng test

`test/test-airdecap-ng-0005.sh` extracts counters by fixed byte offset
(`cut -b 40-`) and hashes them. airdecap-ng refreshes a
`Read N packets...\r` status line once a second, so whenever processing crossed
a one-second boundary that fragment shared a line with the first counter and
shifted the column, changing the hash. The test passed on fast machines and
failed on loaded CI runners. Output is now split on CR before the grep; the
expected hash is unchanged and is now stable whether zero, one or three status
updates print.

*Commit `0676aeb1`.*

### gcc-9 CI job

The dependency step decided whether to install a compiler by checking the C
compiler alone. `gcc-9` and `g++-9` are separate packages and the current
ubuntu-22.04 image ships only the first, so nothing was installed and configure
died with *"A compiler with support for C++11 language features is required"* —
not a C++11 problem, simply a missing compiler. The gcc rows now check the C++
compiler too.

*Commit `edca8a23`.*

---

## Alerts closed without a code change

CodeQL flagged RC4 use in `lib/crypto/arcfour-gcrypt.c`, `lib/crypto/crypto.c`
and `test/cryptounittest/test-cipher-arcfour.c` as a weak cryptographic
algorithm. RC4 is not a choice here: WEP (802.11-1997) and TKIP are defined on
it, and this code is `encrypt_wep()`/`decrypt_wep()` itself. Substituting a
modern cipher would not harden anything, it would stop aircrack-ng from
speaking WEP. Alerts 70, 71 and 75 were dismissed with that rationale rather
than filtered out of the query, so genuinely new weak-crypto use elsewhere is
still reported.

---

## ChangeLog entries

```
* Aircrack-ng: Fix command injection via interface name in the Linux osdep layer (CVE-worthy, CodeQL alert 68)
* Aircrack-ng: Check the full sscanf result when parsing hex wordlist entries
* Aircrack-ng: Require all conversions when reading battery state from /proc/apm
* Aircrack-ng: Stop the /proc/meminfo scan on EOF instead of looping on read errors
* Airmon-ng: Support phys with no bus device, and skip phys without monitor mode
* Airmon-ng: Treat a closed stdin as no rather than looping on the prompt
* Tests: Make the airdecap-ng statistics test immune to the progress status line
* CI: Install g++-N when the runner image only ships gcc-N


kimocoder:
#  * No phy available in VM (AVF/QEMU)

Screenshot_20260804-211939

# * The phy.commit fixes that

Screenshot_20260804-213226

kimocoder:
@ZeroChaos- @Mister-X- 

Hope it helps out, time has given us a few warnings and NL80211 msgs, but I have these, properly tested fixes / improvements, feel free to cherry-pick or merge if you feel it fits.

Another thing, as we now have monitor mode and frame injection on Qualcomm's internal SoC driver QCACLD-3.0, I'll add support for that in a separate PR soon :1st_place_medal: 


[https://www.kali.org/blog/kali-linux-2026-2-release/](https://www.kali.org/blog/kali-linux-2026-2-release/)

- Cheers guys :100: 
Christian B.

kimocoder:
Last commit fixes the "iwconfig" run

Screenshot From 2026-08-04 22-17-22

PR #2724, opened 2026-08-05
Added new CSA-based deauth

singe:
Mathy's work showed that channel switch announcements in beacon frames are a solid deauth primitive for when management frame protection is enabled.

This adds a new `--csa` mode to `aireplay-ng` that functions much like the existing `--deauth` mode.

It waits for a beacon from the target ESSID/BSSID, adds a CSA IE and injects 64 of them per number of attempts chosen.

It uses `mode 0` for an immediate switch, and by defaul channel 14 because that's an invalid channel for all but 20MHz b networks in Japan. The channel is configurable with `--csa-channel`.

When a vulnerable client sees the CSA in the injected beacon frame, it will switch channels. In the case of wpa_supplicant switching to an invalid channel like 14 causes a disconnect and reconnect when it switches back to the correct channel.

This won't work if the network have beacon protection enabled, or if the client ignores unauthenticated beacon-based CSA.

### DIFF
--- a/include/aircrack-ng/support/communications.h
+++ b/include/aircrack-ng/support/communications.h
@@ -99,6 +99,7 @@ struct communication_options
 	int f_iswep;
 
 	uint8_t deauth_rc;
+	uint8_t csa_channel;
 	int r_nbpps;
 	unsigned int r_fctrl;
 	uint8_t r_bssid[6];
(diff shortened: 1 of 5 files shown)


singe:
Unknown

I tested it against hostapd and wpa_supplicant on Debian bookworm with mac80211_hwsim and ieee80211w mandatory on a WPA2 network.

singe:
I saw the broken build tests - realised I forgot to add the header change in my late night commit. Fixed it.

smaddilincoln-ux:
> I saw the broken build tests - realised I forgot to add the header change in my late night commit. Fixed it.

Yoh teach me ho to crack codes

